PR #12, opened 2015-03-14
#27793 Adds list view mode to loader main UI

manneohrstrom:
This change adds a new "list view" mode to the main UI:

image

image

In addition to the above, there are some smaller tweaks as well:

- The delegate widgets have been moved into the loader so that it no longer depends on the standard widgets in the qtwidgets framework. This is a tweak to make it easier for the loader to switch to the v3 qtwidgets framework when that is released as part of the file manager changes.
- General cleanup around rendering of items
- Fixed an issue where actions were retriggered at exit.


app.py
```diff
@@ -18,6 +18,7 @@
 import sys
 import os
 
+
```
jfboismenu-adsk: Typo or PEP8? :)

python/tk_multi_loader/delegate_publish_list.py
```diff
@@ -0,0 +1,343 @@
...
+
+# import the shotgun_model and view modules from the shotgun utils framework
+shotgun_model = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_model")
+shotgun_view = sgtk.platform.import_framework("tk-framework-qtwidgets", "shotgun_view")
+
+from .ui.widget_publish_list import Ui_PublishListWidget
+
+class PublishListWidget(QtGui.QWidget):
+    """
+    Simple list *item* widget which hosts a square thumbnail, header text
```
jfboismenu-adsk: Smells like a copy past from https://github.com/shotgunsoftware/tk-multi-loader2/blob/27793_list_view/python/tk_multi_loader/delegate_publish_history.py#L24

manneohrstrom:
Awesome! Thanks for the feedback! Let me make those tweaks and send some screenshots back for feedback. This was a quick tweak to at least being able to provide an alternative to the thumbnail view, but let me see if I can and get a search box added in at the same time. It would be super useful! :+1:

python/tk_multi_loader/delegate_publish_history.py
```diff
@@ -17,6 +17,107 @@
 shotgun_model = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_model")
 shotgun_view = sgtk.platform.import_framework("tk-framework-qtwidgets", "shotgun_view")
 
+from .ui.widget_publish_history import Ui_PublishHistoryWidget
+
+class PublishHistoryWidget(QtGui.QWidget):
```
alandann: This should really be in it's own file

python/tk_multi_loader/delegate_publish_history.py
```diff
@@ -17,6 +17,107 @@
...
+    def __init__(self, parent):
+        """
+        Constructor
+        
+        :param parent: QT parent object
+        """
+        QtGui.QWidget.__init__(self, parent)
+
+        # make sure this widget isn't shown
+        self.setVisible(False)
```
alandann: I suspect I know why you are having to do this (I think I've just run into the same thing) - we should probably fix this in the base delegate class ;)

python/tk_multi_loader/delegate_publish_history.py
```diff
@@ -17,6 +17,107 @@
...
+        
+        :param header: Header text as string
+        :param body: Body text as string
+        """
+        self.setToolTip("%s<br>%s" % (header, body))        
+        self.ui.header_label.setText(header)
+        self.ui.body_label.setText(body)
+
+    @staticmethod
+    def calculate_size():
```
alandann: Feels like this should be possible by overloading the correct method on the widget e.g. size, sizeHint, minimumSize, minimumSizeHint...  not exactly sure which one though ;)

python/tk_multi_loader/delegate_publish_history.py
```diff
@@ -84,9 +204,7 @@ def _on_before_paint(self, widget, model_index, style_options):
         # v004 (2014-02-21 12:34)
 
         header_str = ""
-
-        if sg_item.get("version_number"):
-            header_str += "<b style='color:#2C93E2'>Version %03d</b>" % sg_item.get("version_number")
+        header_str += "<b style='color:#2C93E2'>Version %03d</b>" % (sg_item.get("version_number") or 0)
```
alandann: is sg_item a dictionary?  if so then sg_item.get("version_number", 0) would be better

python/tk_multi_loader/delegate_publish_history.py
```diff
@@ -95,18 +213,9 @@ def _on_before_paint(self, widget, model_index, style_options):
...
-        else:
-            desc_str = sg_item.get("description")
-        
-        if sg_item.get("created_by") is None:
-            author_str = "Unspecified User"
-        else:
-            author_str = "%s" % sg_item.get("created_by").get("name")
-
+        desc_str = sg_item.get("description") or "No Description Given"         
+        author_str = sg_item.get("created_by").get("name") or "Unspecified User"
```
alandann: sg_item.get("created_by", {}).get("name") would be safer

alandann: actually, I recently had a bug where a field from a shotgun query unexpectedly returned None instead of a dictionary - is it possible in this case that created_by may be None?  In which case splitting across multiple statements and checking each obtained value would be safer...

python/tk_multi_loader/delegate_publish_list.py
```diff
@@ -0,0 +1,380 @@
+# Copyright (c) 2013 Shotgun Software Inc.
```
alandann: Very minor but copyright should be 2015

python/tk_multi_loader/delegate_publish_list.py
```diff
@@ -0,0 +1,380 @@
...
+import datetime
+from .model_latestpublish import SgLatestPublishModel
+
+# import the shotgun_model and view modules from the shotgun utils framework
+shotgun_model = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_model")
+shotgun_view = sgtk.platform.import_framework("tk-framework-qtwidgets", "shotgun_view")
+
+from .ui.widget_publish_list import Ui_PublishListWidget
+
+class PublishListWidget(QtGui.QWidget):
```
alandann: Should be in it's own file

python/tk_multi_loader/delegate_publish_list.py
```diff
@@ -0,0 +1,380 @@
...
+    def __init__(self, parent):
+        """
+        Constructor
+        
+        :param parent: QT parent object
+        """
+        QtGui.QWidget.__init__(self, parent)
+
+        # make sure this widget isn't shown
+        self.setVisible(False)
```
alandann: Same comment as above ;)

python/tk_multi_loader/delegate_publish_list.py
```diff
@@ -0,0 +1,380 @@
...
+        """
+        Called when the associated widget is selected. This method 
+        implements all the setting up and initialization of the widget
+        that needs to take place prior to a user starting to interact with it.
+        
+        :param widget: The widget to operate on (created via _create_widget)
+        :param model_index: The model index to operate on
+        :param style_options: QT style options
+        """
+        # first set up the basic
```
alandann: incomplete comment - doesn't make much sense!

python/tk_multi_loader/delegate_publish_list.py
```diff
@@ -0,0 +1,380 @@
...
+        elif isinstance(field_value, list):
+            # this is a list of some sort. Loop over all elements and extract a comma separated list.
+            # this can be a multi link field but also a field like a tags field or a non-entity link type field.
+            formatted_values = []
+            formatted_types = set()
+            
+            for v in field_value:
+                if isinstance(v, dict) and "name" in v and "type" in v:
+                    # This is a link field
+                    if v.get("name"):
```
alandann: this could be more readable and only do the dictionary look-up once, e.g.

name = v["name"]
if name:
    formatted_values.append(name)
    ...

python/tk_multi_loader/delegate_publish_list.py
```diff
@@ -0,0 +1,380 @@
...
+                        formatted_values.append(v.get("name"))
+                        formatted_types.add(v["type"])
+                else:
+                    formatted_values.append(str(v))
+            
+            if len(formatted_types) == 0:
```
alandann: This if/else isn't needed as types = ", ".join(formatted_types) works even if the set is empty (it will return an empty string)

python/tk_multi_loader/delegate_publish_list.py
```diff
@@ -0,0 +1,380 @@
...
+                main_text += "<b>%s</b> <b style='color:#2C93E2'>%s</b>" % (entity_link["type"], entity_link["name"])
+
+            if sg_data.get("task") is not None:
+                main_text += ", Task %s" % sg_data["task"]["name"]
+
+        # Version 012 by John Smith at 2014-02-23 10:34            
+        created_unixtime = sg_data.get("created_at") or 0
+        date_str = datetime.datetime.fromtimestamp(created_unixtime).strftime('%Y-%m-%d %H:%M')
+        small_text = "Version %03d by %s at %s" % (sg_data.get("version_number"), 
+                                                      sg_data.get("created_by").get("name"),
```
alandann: same here - are you sure "created_by" can't be None?

python/tk_multi_loader/delegate_publish_thumb.py
```diff
@@ -16,40 +17,176 @@
 shotgun_model = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_model")
 shotgun_view = sgtk.platform.import_framework("tk-framework-qtwidgets", "shotgun_view")
 
+from .ui.widget_publish_thumb import Ui_PublishThumbWidget
 
-class SgPublishDelegate(shotgun_view.WidgetDelegate):
+
+class PublishThumbWidget(QtGui.QWidget):
```
alandann: Should be in it's own file

python/tk_multi_loader/delegate_publish_thumb.py
```diff
@@ -16,40 +17,176 @@
...
+        # make sure this widget isn't shown
+        self.setVisible(False)
+        
+        # set up the UI
+        self.ui = Ui_PublishThumbWidget() 
+        self.ui.setupUi(self)
+        
+        # set up an event filter to ensure that the thumbnails
+        # are scaled in a square fashion.
+        filter = ResizeEventFilter(self.ui.thumbnail)
```
alandann: ResizeEventFilter is a really ambiguous name - maybe make it more specific or move it inside this class as a protected sub-class?

python/tk_multi_loader/delegate_publish_thumb.py
```diff
@@ -165,26 +368,49 @@ def _on_before_paint(self, widget, model_index, style_options):
...
-        return shotgun_view.ThumbWidget.calculate_size(scale_factor)
+        return PublishThumbWidget.calculate_size(scale_factor)
+
+
 
+##################################################################################################
+# utility classes
+
+
+class ResizeEventFilter(QtCore.QObject):
```
alandann: see comment above where this is used

python/tk_multi_loader/delegate_publish_thumb.py
```diff
@@ -165,26 +368,49 @@ def _on_before_paint(self, widget, model_index, style_options):
...
+    """
+    Event filter which emits a resized signal whenever
+    the monitored widget resizes
+    """
+    resized = QtCore.Signal()
+
+    def eventFilter(self, obj, event):
```
alandann: Missing doc-string

python/tk_multi_loader/dialog.py
```diff
@@ -363,6 +379,46 @@ def _on_history_double_clicked(self, model_index):
...
+        self._set_main_view_mode(self.MAIN_VIEW_LIST)
+
+    def _set_main_view_mode(self, mode):
+        """
+        Sets up the view mode for the main view.
+        
+        :param mode: either MAIN_VIEW_LIST or MAIN_VIEW_THUMB
+        """
+        if mode == self.MAIN_VIEW_LIST:
+            self.ui.list_mode.setIcon(QtGui.QIcon(QtGui.QPixmap(":/res/mode_switch_card_active.png")))
```
alandann: not major but this could be done through the style sheet instead (by setting the icons on the sub-states)

python/tk_multi_loader/dialog.py
```diff
@@ -363,6 +379,46 @@ def _on_history_double_clicked(self, model_index):
...
+            self.ui.list_mode.setChecked(True)
+            self.ui.thumbnail_mode.setIcon(QtGui.QIcon(QtGui.QPixmap(":/res/mode_switch_thumb.png")))
+            self.ui.thumbnail_mode.setChecked(False)
```
alandann: Using radio buttons would also reduce the code here

python/tk_multi_loader/dialog.py
```diff
@@ -363,6 +379,46 @@ def _on_history_double_clicked(self, model_index):
...
+            self.ui.list_mode.setChecked(False)
+            self.ui.thumbnail_mode.setIcon(QtGui.QIcon(QtGui.QPixmap(":/res/mode_switch_thumb_active.png")))
+            self.ui.thumbnail_mode.setChecked(True)
+            self.ui.publish_view.setViewMode(QtGui.QListView.IconMode)
+            self.ui.publish_view.setItemDelegate(self._publish_thumb_delegate)
+            
+        else:
+            raise TankError("Undefined view mode!") 
+
+        self.ui.publish_view.selectionModel().clear()
```
alandann: the behaviour that switching views clears the selection seems odd to me

alandann:
cr done - a few comments but nothing major

python/tk_multi_loader/delegate_publish_history.py
manneohrstrom: This is how it started out - that was my gut feeling too - but it actually felt better once i had moved it inside this file - because the delegate and the widget are so related. I can move it out if you feel strongly about it, but in my opinion this would be a case where a good general rule of thumb is deliberately broken for improved readability! :)

python/tk_multi_loader/delegate_publish_history.py
manneohrstrom: This code is really old (i just moved it from the qtwidgets code as part of this pull request!) so I don't even fully remember why it's done this way. :) Do we have a list of shotgunutils/qtwidgets v3 stuff we should be doing? If not, let's huddle and get that list going - great opportunity here to fix some stuff in v2 which weren't ideal!

python/tk_multi_loader/delegate_publish_history.py
manneohrstrom: Another one for the v3 actions list!

python/tk_multi_loader/delegate_publish_history.py
manneohrstrom: Because it is shotgun data, it is common that the dictionary data can be returned as None - this is why this `sg_data.get("field_name") or "default"` pattern is repeated everywhere. I didn't comment on it because it is something we do all the time. Does that make sense?

python/tk_multi_loader/delegate_publish_history.py
manneohrstrom: I think `sg_item.get("created_by", {}).get("name")` would give a false sense of security, because it would be returned as `None` from shotgun if the value isn't defined but the field is requested! The code does make the assumption that `created_by` always exist - I think this is set for every record, whatever you do in Shotgun. I thought it was a safe assumption to make because created_by is such a special field, like creation_time, updated_by etc (I think they are called audit fields!). Does that make sense? Happy to put in additional checks, but only if you absolutely feel we need to!

python/tk_multi_loader/delegate_publish_list.py
manneohrstrom: see above :)

python/tk_multi_loader/delegate_publish_list.py
manneohrstrom: see above :)

python/tk_multi_loader/delegate_publish_list.py
manneohrstrom: will fix!

python/tk_multi_loader/delegate_publish_list.py
manneohrstrom: will fix!

python/tk_multi_loader/delegate_publish_list.py
manneohrstrom: good point - will fix!

python/tk_multi_loader/delegate_publish_list.py
manneohrstrom: good point will fix! I think previously, there was something else going on in that if/else clause, which later on got removed, leaving us with this. Nice to clean up! :)

python/tk_multi_loader/delegate_publish_list.py
manneohrstrom: see above! :)

python/tk_multi_loader/delegate_publish_thumb.py
manneohrstrom: see above! :)

python/tk_multi_loader/delegate_publish_thumb.py
manneohrstrom: I thought it was enough to keep it in the same file to indicate that this class was internal to the .py file and wasn't meant to be used anywhere else! I don't generally like nested classes but it's a good point - i'll shift it round :)

python/tk_multi_loader/delegate_publish_thumb.py
manneohrstrom: will fix

python/tk_multi_loader/dialog.py
manneohrstrom: can you make radio buttons look the same? The current setup is pretty clunky, i'll see if i can simplify it with some style sheeting and will try the radio buttons...

python/tk_multi_loader/dialog.py
manneohrstrom: It's a delegate weirdness thing. i'll add a comment and will add it to the v3 doc for the libraries :)

manneohrstrom:
Great comments @alandann will make the changes - left some of them as questions to check what you think. Will also make the adjustments that rodeo suggested for the formatting and report back!

python/tk_multi_loader/dialog.py
manneohrstrom: here's a recipe (thanks @alandann)

```
#grid_radio_btn::indicator {
width: 26;
height: 24;
}

#grid_radio_btn::indicator::unchecked {
image: url(:/tk-multi-workfiles/grid_view_unchecked.png);
}

#grid_radio_btn::indicator::unchecked::hover {
image: url(:/tk-multi-workfiles/grid_view_unchecked_hover.png);
}

#grid_radio_btn::indicator::checked {
    image: url(:/tk-multi-workfiles/grid_view_checked.png);
}
```

python/tk_multi_loader/delegate_publish_history.py
jfboismenu-adsk: Could we put this on the technical debt list and see if we can do something for 1.0? This is so error prone than even a Toolkit veteran got it wrong. Not a good sign.

python/tk_multi_loader/delegate_publish_history.py
jfboismenu-adsk: If the intent is that creater_by should always be there, I think you should use [], otherwise it looks like you might have forgot to add a default value. Plus, if the create_by field is ever missing, the "KeyError: created_by" exception string is more telling than a "NoneType has no attribute get'.

python/tk_multi_loader/delegate_publish_list.py
jfboismenu-adsk: See above for my comment too! :)

python/tk_multi_loader/delegate_publish_history.py
manneohrstrom: I don't think this is technical debt – isn't it just code? A simple, pragmatic way to have it display "Version 0" instead of "Version None"?

jfboismenu-adsk: Yes, but what I getting at is that I would have thought like Alan that sg_data.get("version_number", 0) would have worked. It's confusing to have None stored in the dictionary for that entry. That's my take on it anyway.

python/tk_multi_loader/delegate_publish_history.py
manneohrstrom: good point!

python/tk_multi_loader/delegate_publish_history.py
manneohrstrom: It's the shotgun API basically - you ask it "find shots for me and return values name, in_frame and description" - and the data it returns is None for all those fields where no value has been set; 

```
{"type": "Shot", "id": 1234, "description": "Foo bar", "in_frame": None }
```

manneohrstrom:
Addressed Code review and tweaked the UI based on @rodeo 's suggestions:

image

PR #15, opened 2015-07-15
Ticket #30252 - Show task info in the lists delegate.

robblau:
Show the task info on the list view in the loader:

shotgun__loader

Always show it since we have the space and the logic about when it is shown or not in the Thumbnail view can be a little subtle and non-obvious.


manneohrstrom:
Looks good!

PR #16, opened 2015-09-21
Add status field to preloaded fields

jesseemond-adsk:
This adds the `sg_status_list` to the list of preloaded fields of published files before passing them to the `filter_publishes` hook.

This simplifies the creation of a hook that hides publishes under a certain status (e.g. filtering out publishes that are under a custom *deprecated* status) while only adding one standard preloaded field.

### DIFF
--- a/python/tk_multi_loader/constants.py
+++ b/python/tk_multi_loader/constants.py
@@ -1,0 +1,34 @@
+# Copyright (c) 2015 Shotgun Software Inc.
+# 
+# CONFIDENTIAL AND PROPRIETARY
+# 
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+# Source Code License included in this distribution package. See LICENSE.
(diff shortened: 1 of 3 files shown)


manneohrstrom:
Jesse - nice one! :+1: :+1: 

One very related however more general suggestion/idea for the loader while you are already in there:

We actually have that same list of fields in two places:

- https://github.com/shotgunsoftware/tk-multi-loader2/blob/master/python/tk_multi_loader/model_publishhistory.py#L63

- https://github.com/shotgunsoftware/tk-multi-loader2/blob/master/python/tk_multi_loader/model_latestpublish.py#L233

Could you add something like a constants file and break those lists out into a shared place (I am thinking perhaps like we do here: https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/platform/constants.py )? Something we should have done a long time ago but haven't -- if you could do it now while you are in that part of the code anyways, that would be great!

Thanks!

PR #18, opened 2015-09-25
#31343 Apply 'filter_publishes' in Version History

jesseemond-adsk:
The `filter_publishes` hook is only applied to the main published files view and not to the version history (in the details section). This behaviour is not really intuitive as one might expect the filter to affect the entire list of publishes, no matter where or how it is represented. This PR makes it so that the hook is also called for the version history.

Done by factoring out the filtering logic to a `utils.py` function, to reduce redundancy.

Also took the opportunity to make `utils.py` PEP8 compliant and fix (what I think are) various minor typos.

### DIFF
--- a/python/tk_multi_loader/utils.py
+++ b/python/tk_multi_loader/utils.py
@@ -215,7 +215,7 @@ def create_overlayed_publish_thumbnail(image):
         painter.setRenderHint(QtGui.QPainter.Antialiasing)
         painter.setBrush(brush)
 
-        # figure out the offset height wise in order to center the thumb
+        # figure out the vertical offset in order to center the thumb
         height_difference = CANVAS_HEIGHT - thumb_scaled.height()
         width_difference = CANVAS_WIDTH - thumb_scaled.width()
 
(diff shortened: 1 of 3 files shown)


python/tk_multi_loader/utils.py
```diff
@@ -193,48 +193,86 @@ def create_overlayed_publish_thumbnail(image):
...
+            hook_publish_list = []
+
+        # split back out publishes:
+        sg_data_list = []
+        for item in hook_publish_list:
+            sg_data = item.get("sg_publish")
+            if sg_data:
+                sg_data_list.append(sg_data)
+
+    except Exception:
```
manneohrstrom: Nitty gritty: I don't think the `Exception` bit is needed.

python/tk_multi_loader/utils.py
```diff
@@ -193,48 +193,86 @@ def create_overlayed_publish_thumbnail(image):
...
     return base_image
-    
+
+
+def filter_publishes(app, sg_data_list):
+    """
+    Filters a list of shotgun published files based on the filter_publishes
+    hook.
+
+    :param app:           application that has the hook.
```
manneohrstrom: minor:  I guess this should say app rather than application :)

manneohrstrom:
Two super minor comments! good to go!

PR #19, opened 2015-10-02
For #31344, crash on empty created_by field.

jfboismenu-adsk:
We're now testing for the created_by field to be set before trying to get the name from the field.


manneohrstrom:
CR looks good! :)

PR #23, opened 2016-01-19
#33333 Added support for loading published alembic cache publishes as houdini alembic/alembicarchive nodes

josh-t:
This is a very quick and simple implementation for importing published alembic caches as houdini alembic and alembicarchive nodes. As with the other engines, this implementation can serve as a basic starting point for a more robust implementation out in the world.

### DIFF
--- a/hooks/tk-houdini_actions.py
+++ b/hooks/tk-houdini_actions.py
@@ -142,6 +142,12 @@ class HoudiniActions(HookBaseClass):
 
     ##############################################################################################################
     def _import(self, path, sg_publish_data):
+        """Import the supplied path as a geo/alembic sop.
+        
+        :param str path: The path to the file to import.
+        :param dict sg_publish_data: The publish data for the supplied path.
+        
+        """
 
         import hou
         app = self.parent
@@ -149,9 +155,15 @@ class HoudiniActions(HookBaseClass):
         name = sg_publish_data.get("name", "alembic")
         path = self.get_publish_path(sg_publish_data)
 
-        obj_context = hou.node("/obj")
+        obj_context = _get_current_context("/obj")
         app.log_debug("Creating geo: %s" % (name,))
-        geo_node = obj_context.createNode("geo", name)
+
+        try:
+            geo_node = obj_context.createNode("geo", name)
+        except hou.OperationFailed:
+            # failed to create the node in this context, create at top-level
+            obj_context = hou.node("/obj")
+            geo_node = obj_context.createNode("geo", name)
 
         # delete the default nodes created in the geo
         for child in geo_node.children():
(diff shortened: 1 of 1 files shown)


hooks/tk-houdini_actions.py
```diff
@@ -116,4 +137,46 @@ def _merge(self, path, sg_publish_data):
...
+    ##############################################################################################################
+    def _import(self, path, sg_publish_data):
+
+        import hou
+        app = self.parent
+
+        name = sg_publish_data.get("name", "alembic")
+        path = self.get_publish_path(sg_publish_data)
+
+        obj_context = hou.node("/obj")
```
robblau: If the current context could contain a geo node, seems like we should create it there (right?).
That way if you are deep in a subnetwork you could import the cache there.

hooks/tk-houdini_actions.py
```diff
@@ -116,4 +137,46 @@ def _merge(self, path, sg_publish_data):
...
+    ##############################################################################################################
+    def _import_archive(self, path, sg_publish_data):
+
+        import hou
+        app = self.parent
+
+        name = sg_publish_data.get("name", "alembicarchive")
+        path = self.get_publish_path(sg_publish_data)
+
+        obj_context = hou.node("/obj")
```
robblau: If the current context could contain a geo node, seems like we should create it there (right?).
That way if you are deep in a subnetwork you could import the cache there.

hooks/tk-houdini_actions.py
josh-t: I modified the code to first attempt to retrieve the current obj context and create the node there. If it can't create it there it falls back to top-level "/obj". Will resubmit for CR so you can see if what I did makes sense.

hooks/tk-houdini_actions.py
josh-t: I modified the code to first attempt to retrieve the current obj context and create the node there. If it can't create it there it falls back to top-level "/obj". Will resubmit for CR so you can see if what I did makes sense.

hooks/tk-houdini_actions.py
```diff
@@ -116,4 +137,147 @@ def _merge(self, path, sg_publish_data):
...
+        """
+
+        import hou
+        app = self.parent
+
+        name = sg_publish_data.get("name", "alembic")
+        path = self.get_publish_path(sg_publish_data)
+
+        obj_context = _get_current_context("/obj")
+        app.log_debug("Creating geo: %s" % (name,))
```
robblau: Would be good to mention the path to the context it'll be created in.  Just for debugging help.

hooks/tk-houdini_actions.py
```diff
@@ -116,4 +137,147 @@ def _merge(self, path, sg_publish_data):
...
+            obj_context = hou.node("/obj")
+            geo_node = obj_context.createNode("geo", name)
+
+        # delete the default nodes created in the geo
+        for child in geo_node.children():
+            child.destroy()
+
+        app.log_debug("Creating alembic sop: %s" % (name,))
+        alembic_sop = geo_node.createNode("alembic", name)
+        app.log_debug("Setting alembic sop path: %s" % (path,))
```
robblau: Could these debug statements be consolidated so there wouldn't be so many lines?  Not a big deal, but it seems a little redundant.

hooks/tk-houdini_actions.py
josh-t: done.

hooks/tk-houdini_actions.py
josh-t: done.

PR #28, opened 2016-10-17
Multi-Selection in the Shotgun Toolkit Loader 

jfboismenu:
**Purpose of the PR / What does this do?**
- adds multi-selection support to the loader, which allows to execute a single action on multiple publishes
- fixes tooltip so widget doesn't need to be selected to have one.
- thumbnail scale slider is hidden when in list mode
- when more than one item is selected, history view is disable and visual status is updated.

screen shot 2016-11-03 at 11 10 57

screen shot 2016-10-21 at 14 36 28

**Overview of the changes**
- Refactored the `delegate_publish_list.py` and `delegate_publish_thumb.py` file since there was a lot of copy pasted code in here. That code has been refactored into `delegate_publish.py`.
- Updated the selection code to be able to pick more than one item
- Added code that remove actions that are not common to every publishes
- Updated the Omnigraffle resource and generated a new image for the history pane when there is more than one publish selected. Note that I've had to regenerate another resource (No publish selected!) so that toggling between the two images will not cause realignment of the text and icon.
- Introduced a backwards compatible hook method that allows to capture the entire process of executing the action on every items in `execute_action_on_selection` so that hook writers can own what it means to apply an action on multiple items at a time.
- Updated the builtin hooks for every DCCs.
- Fixed tooltip generation so they are generated for all items in the model and not just the currently selected item.

**Where should the reviewer start looking at?**

Great care should be taken when reviewing `loader_action_manager.py` which is where we are detecting which actions are common to every publishes. Also, the refactoring of common functionality from `delegate_publish_thumb.py` and `delegate_publish_list.py` into `publish_delegate.py` should be looked at closely.


hooks/tk-3dsmax_actions.py
```diff
@@ -78,7 +78,39 @@ def generate_actions(self, sg_publish_data, actions, ui_area):
                                       "description": "This will insert a reference to this file into the current scene."} )        
         
         return action_instances
-                
+
+    def execute_action_on_selection(self, name, action_params):
```
jfboismenu: This is the backwards compatible code that is added for every DCC.

python/tk_multi_loader/delegate_publish_list.py
```diff
@@ -30,86 +31,17 @@ def __init__(self, parent):
         
         :param parent: QT parent object
         """
-        QtGui.QWidget.__init__(self, parent)
+        PublishWidget.__init__(self, Ui_PublishListWidget, parent)
 
-        # make sure this widget isn't shown
```
jfboismenu: All removed code has gone into the base class.

python/tk_multi_loader/delegate_publish_list.py
```diff
@@ -30,86 +31,17 @@ def __init__(self, parent):
...
-        Set a thumbnail given the current pixmap.
-        The thumbnail must be 512x400px or similar aspect
-        not to be squeezed.
-        
-        :param pixmap: pixmap object to use
-        """
-        self.ui.thumbnail.setPixmap(pixmap)
-            
-    def set_text(self, large_text, small_text, tooltip):
+    def set_text(self, large_text, small_text):
```
jfboismenu: Tooltips are now handled by the model.

python/tk_multi_loader/delegate_publish_thumb.py
```diff
@@ -34,81 +34,17 @@ def __init__(self, parent):
         
         :param parent: QT parent object
         """
-        QtGui.QWidget.__init__(self, parent)
+        PublishWidget.__init__(self, Ui_PublishThumbWidget, parent)
 
-        # make sure this widget isn't shown
```
jfboismenu: All removed code has gone into the base class.

python/tk_multi_loader/delegate_publish_thumb.py
```diff
@@ -34,81 +34,17 @@ def __init__(self, parent):
...
-        """
-        Set a thumbnail given the current pixmap.
-        The pixmap must be 512x400 aspect ratio or it will appear squeezed
-        
-        :param pixmap: pixmap object to use
-        """
-        self.ui.thumbnail.setPixmap(pixmap)
-    
-    def set_text(self, header, body, tooltip):
+    def set_text(self, header, body):
```
jfboismenu: Tooltips are now handled by the model.

python/tk_multi_loader/loader_action_manager.py
```diff
@@ -105,40 +81,133 @@ def get_actions_for_publish(self, sg_data, ui_area):
...
         except Exception:
             self._app.log_exception("Could not execute generate_actions hook.")
-            
-            
-            
-        # create QActions
+
+        return action_defs_dict
+
+    def get_actions_for_publishes(self, sg_data_list, ui_area):
```
jfboismenu: This is the new entry point on the action manager to get actions for a given selection.

python/tk_multi_loader/action_manager.py
```diff
@@ -27,12 +27,11 @@ def __init__(self):
         """
         QtCore.QObject.__init__(self)
 
-    def get_actions_for_publish(self, sg_data, ui_area):
+    def get_actions_for_publishes(self, sg_data, ui_area):
```
jfboismenu: Other parts of the app, like the history view, won't support multiple selection for now for lack of a useful workflow so this is still required.

resources/resources.qrc
```diff
@@ -10,6 +10,7 @@
     <file>play_icon.png</file>
     <file>no_publishes_found.png</file>
     <file>no_item_selected_512x400.png</file>
+    <file>too_many_publishes_512x400.png</file>
     <file>folder_512x400.png</file>
     <file>loading_512x400.png</file>
     <file>loading_100x100.png</file>
```
jfboismenu: Can't comment on image below, so commenting here. The light gray text reads:

Please select a Publish to see more details.

heavyimage:
Hey JF!

Forgive me if this is jumping the gun here -- I'm not sure if this code is intended for use in production quite yet, however @manneohrstrom pointed me at this code because he knows we're really excited about loader multi-select.  By adding rodeo's fork of loader2 to my local repo and merging your changes, I was able to get this to work in our environment which is super duper mega exciting.  However, I noticed a few regressions I thought I'd bring to your attention here -- apologies if this is uninvited or these issues are already known.

* The "Actions" button on each published file seems to now be broken -- you can click it, but no popup appears.  While the right-click functionality makes sense for performing an action on a multiselection, I believe the actions button should still function for one-offs.

* The order of the actions has changed.  This is problematic because we've made sure that the first action for each published file type is the action which loads that data into our scenes so that 1) double-clicking works and 2) that the most common action is at the top of the list.  Now that that order has changed, double clicking does something arbitrary and loading an item into your scene is no longer at the top of the list.

* The change below has stopped filtering from working so now I'm seeing all published file types in the loader as opposed to the subset I've decided to include:

@@ -805,9 +820,7 @@ class AppDialog(QtGui.QWidget):
         # go through and figure out which checkboxes are clicked and then
         # update the publish proxy model so that only items of that type
         # is displayed
-        sg_type_ids = self._publish_type_model.get_selected_types()
         show_folders = self._publish_type_model.get_show_folders()
-        self._publish_proxy_model.set_filter_by_type_ids(sg_type_ids, show_folders)

That's what I've noticed so far -- again, I hope this is useful and not annoying!
Jesse

jfboismenu:
Hey Jesse, no, your feedback is great! I'll will fix this as soon as I can.

jfboismenu:
All issues mentioned above by @heavyimage have been fixed.

heavyimage:
Wow!  That's pretty complete looking!  I'll try this out and report back!

python/tk_multi_loader/dialog.py
```diff
@@ -549,19 +566,34 @@ def __set_publish_ui_visibility(is_publish):
             self.ui.detail_actions_btn.setVisible(is_publish)
             self.ui.detail_playback_btn.setVisible(is_publish)
 
+        def __clear_publish_history(pixmap):
+            self._publish_history_model.clear()
```
manneohrstrom: docstring

python/tk_multi_loader/model_latestpublish.py
```diff
@@ -198,6 +198,38 @@ def async_refresh(self):
         """
         self._refresh_data()
 
+    def _set_tooltip(self, item, sg_item):
```
manneohrstrom: nice!

hooks/tk-3dsmaxplus_actions.py
```diff
@@ -79,7 +79,39 @@ def generate_actions(self, sg_publish_data, actions, ui_area):
                                       "description": "This will insert a reference to this file into the current scene."} )        
         
         return action_instances
-                
+
+    def execute_action_on_selection(self, name, action_params):
```
jfboismenu: True! I'll go for `execute_multiple_actions`, which is generic enough and less error prone than `execute_actions`.

hooks/tk-3dsmaxplus_actions.py
```diff
@@ -79,7 +79,39 @@ def generate_actions(self, sg_publish_data, actions, ui_area):
...
+        The default implementation dispatches each item from ``action_params`` to
+        the ``execute_action`` method.
+
+        The ``action_params`` will take the following layout:
+
+        .. code-block::
+            [
+                (
+                    {"type": "PublishedFile", "id": 3, ...},
+                    # Value returned in the "params" field of "generate_actions" return value>
```
jfboismenu: Good point.

hooks/tk-shell_actions.py
```diff
@@ -0,0 +1,144 @@
...
+"""
+Hook that loads defines all the available actions, broken down by publish type.
+"""
+import pprint
+import sgtk
+
+HookBaseClass = sgtk.get_hook_baseclass()
+
+
+class ShellActions(HookBaseClass):
```
jfboismenu: Good point. I've renamed the actions to be named Debug Action 1 and so on. I'd rather not have them be hidden behind a debug descriptor however. I think it is implied that these apps make no sense outside a DCC and as such the hooks are there for testing. I'll ask @robblau about it.

python/tk_multi_loader/delegate_publish_thumb.py
```diff
@@ -123,33 +59,7 @@ def calculate_size(scale_factor):
...
-    def set_sub_items_mode(self, enabled):
-        """
-        Enables rendering of cells in to work with the sub items
-        mode, where the result list will contain items from several
-        different folder levels.
-
-        :param enabled: True if subitems mode is enabled, false if not
-        """
-        self._sub_items_mode = enabled
+class SgPublishThumbDelegate(PublishDelegate):
```
jfboismenu: Thanks, I lot some of these during refactoring.

jfboismenu:
I totally agree with the Too Many Selected. I wasn't in love with the wording and I think you've phrase it perfectly. I'll update the resources accordingly.

PR #31, opened 2016-12-02
#39476: 'Show in the file system' available only when a folder widget has some paths.

fpainchaud:
Add the _Show in the file system_ action to a folder widget only when there are some paths associated with its Shotgun entity. This is done at runtime when the folder widget is selected.

Note that the paths are retrieved with https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/api.py#L593 that uses the path cache to look up all paths associated with the entity.

### DIFF
--- a/python/tk_multi_loader/loader_action_manager.py
+++ b/python/tk_multi_loader/loader_action_manager.py
@@ -268,18 +268,33 @@ class LoaderActionManager(ActionManager):
         
     def get_actions_for_folder(self, sg_data):
         """
-        Returns a list of actions for a folder object.
+        Returns a list of actions for a folder widget.
+
+        This method is called at runtime when a folder widget with Shotgun data is selected.
+
+        :param sg_data: Standard Shotgun entity dictionary with keys type, id and name.
+        :return: List of QAction instances.
         """
-        fs = QtGui.QAction("Show in the file system", None)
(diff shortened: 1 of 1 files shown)


python/tk_multi_loader/loader_action_manager.py
```diff
@@ -268,18 +268,33 @@ def has_actions(self, publish_type):
...
+        :return: List of QAction instances.
         """
-        fs = QtGui.QAction("Show in the file system", None)
-        fs.triggered[()].connect(lambda f=sg_data: self._show_in_fs(f))
-        
+
+        actions = []
+
+        # Find paths associated with the Shotgun entity.
+        paths = self._app.sgtk.paths_from_entity(sg_data["type"], sg_data["id"])
```
manneohrstrom: my only concern here is that this wouldn't be performant enough if you are rendering a view of say 500 shots or more. Can you do a quick test?

fpainchaud: It must be noted that method `paths_from_entity` returns paths stored in Shotgun as `FilesystemLocation` entities: https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/api.py#L573

With my test site, the apps I used and the scripts I wrote, I could not build a proper test case in a reasonable amount of time. I will ask QA to keep an eye on the performance concern.

PR #32, opened 2016-12-14
#39260 filter support for new shotgun model

manneohrstrom:
This updates the loader to work correctly with the new optimized shotgun model that defer loads things into memory.

- the filtering is 'capped' so that it doesn't kick in until you have typed 2 characters.
- the entire tree is loaded into memory at this point in order for the proxy model to operate.

### DIFF
--- a/python/tk_multi_loader/constants.py
+++ b/python/tk_multi_loader/constants.py
@@ -35,4 +35,4 @@ PUBLISHED_FILES_FIELDS = ["name",
 
 # left hand side tree view search only kicks in
 # after a certain number have been typed in.
-TREE_SEARCH_TRIGGER_LENGTH = 1
+TREE_SEARCH_TRIGGER_LENGTH = 2
--- a/python/tk_multi_loader/constants.py
+++ b/python/tk_multi_loader/constants.py
@@ -32,3 +32,7 @@ PUBLISHED_FILES_FIELDS = ["name",
                           "version.Version.sg_status_list",
                           "created_by.HumanUser.image"
                           ]
+
+# left hand side tree view search only kicks in
+# after a certain number have been typed in.
+TREE_SEARCH_TRIGGER_LENGTH = 1
--- a/python/tk_multi_loader/dialog.py
+++ b/python/tk_multi_loader/dialog.py
@@ -1280,7 +1280,7 @@ class AppDialog(QtGui.QWidget):
         # change UI decorations based on new pattern.
         # for performance, make sure filtering only kicks in
         # once we have typed at least one character
-        if pattern and len(pattern) > constants.TREE_SEARCH_TRIGGER_LENGTH:
+        if pattern and len(pattern) >= constants.TREE_SEARCH_TRIGGER_LENGTH:
             # indicate with a blue border that a search is active
             tree_view.setStyleSheet("""QTreeView { border-width: 3px;
                                                    border-style: solid;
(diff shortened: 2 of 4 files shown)


python/tk_multi_loader/constants.py
```diff
@@ -32,3 +32,7 @@
                           "version.Version.sg_status_list",
                           "created_by.HumanUser.image"
                           ]
+
+# left hand side tree view search only kicks in
+# after a certain number have been typed in.
+TREE_SEARCH_TRIGGER_LENGTH = 1
```
josh-t: Based on the PR summary, I would expect this to value to be `2`.

python/tk_multi_loader/dialog.py
```diff
@@ -1276,7 +1278,9 @@ def _on_search_text_changed(self, pattern, tree_view, proxy_model):
         proxy_model.setFilterFixedString(pattern)
 
         # change UI decorations based on new pattern.
-        if pattern and len(pattern) > 0:
+        # for performance, make sure filtering only kicks in
+        # once we have typed at least one character
+        if pattern and len(pattern) > constants.TREE_SEARCH_TRIGGER_LENGTH:
```
josh-t: Suggest making the value of `TREE_SEARCH_TRIGGER_LENGTH` be `2` and making this `>=` just for clarity. i.e. the tree search isn't triggered until the length is 2 or greater.

python/tk_multi_loader/dialog.py
```diff
@@ -1276,7 +1278,9 @@ def _on_search_text_changed(self, pattern, tree_view, proxy_model):
         proxy_model.setFilterFixedString(pattern)
 
         # change UI decorations based on new pattern.
-        if pattern and len(pattern) > 0:
+        # for performance, make sure filtering only kicks in
+        # once we have typed at least one character
```
josh-t: should be: typed at least **two** character**s**

python/tk_multi_loader/proxymodel_entity.py
```diff
@@ -86,8 +88,29 @@ def setFilterFixedString(self, pattern):
         self._cache_hits = 0
         self._cache = {}
 
-        # call base class
-        return QtGui.QSortFilterProxyModel.setFilterFixedString(self, pattern)
+        if len(pattern) > constants.TREE_SEARCH_TRIGGER_LENGTH:
```
josh-t: Same comment here. `>=`.

python/tk_multi_loader/proxymodel_entity.py
```diff
@@ -86,8 +88,29 @@ def setFilterFixedString(self, pattern):
...
+            # we have a search filter that is longer than one character.
+            # start filtering. Before we can filter, ensure that the entire
+            # data set is loaded in the tree.
+
+            def __fetch_more_r(index):
+                if self.canFetchMore(index):
+                    self.fetchMore(index)
```
josh-t: Same comment as other PR here. I'm assuming it's fine but this requires `fetchMore` to be synchronous in order for `rowCount` to be reliable right? Are we ok to make that assumption here?

python/tk_multi_loader/dialog.py
manneohrstrom: done

python/tk_multi_loader/dialog.py
manneohrstrom: done!

python/tk_multi_loader/proxymodel_entity.py
manneohrstrom: done!

python/tk_multi_loader/proxymodel_entity.py
manneohrstrom: fixed!

manneohrstrom:
ready for CR

PR #34, opened 2017-01-04
#38537: Adds tk-adobecc support.

thebeeland:
Note that this also deprecates support for tk-photoshop at the same time. This will not be merged until release of tk-adobecc and subsequent deprecation of tk-photoshop.


python/tk_multi_loader/loader_action_manager.py
```diff
@@ -96,8 +96,8 @@ def _get_actions_for_publish(self, sg_data, ui_area):
                                                         sg_publish_data=sg_data,
                                                         actions=actions,
                                                         ui_area=ui_area_str)
-        except Exception:
-            self._app.log_exception("Could not execute generate_actions hook.")
+        except Exception, e:
+            self._app.log_exception("Could not execute generate_actions hook: %s" % e)
```
manneohrstrom: Heads up that this will include the exception message twice.

The following code:
```python
e = sgtk.platform.current_engine()

try:
    print "foo" + xyz
except Exception, ex:
    e.log_exception("Could not execute actions hook: %s" % ex)
```

Gives the following output:
```
# Error: Shotgun: tk-maya: Could not execute actions hook: name 'xyz' is not defined
Traceback (most recent call last):
  File "<maya console>", line 8, in <module>
NameError: name 'xyz' is not defined # 
```

The original code seems better to me:

```
e = sgtk.platform.current_engine()
try:
    print "foo" + xyz
except Exception:
    e.log_exception("Could not execute actions hook")
    
<Sgtk Engine 0x15be7d810: tk-maya, env: site>
# Error: Shotgun: tk-maya: Could not execute actions hook
Traceback (most recent call last):
  File "<maya console>", line 8, in <module>
NameError: name 'xyz' is not defined # 
```

hooks/tk-adobecc_actions.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def _open_file(self, path, sg_publish_data):
+        """
+        Import contents of the given file into the scene.
+        
+        :param path: Path to file.
+        :param sg_publish_data: Shotgun data dictionary with all the standard
+                                publish fields.
+        """
+        path = "/".join(path.split(os.path.sep))
```
josh-t: Interesting. Does adobe always expect `/`?

thebeeland: More likely is the inability/difficulty in properly escaping a backslash due to the number of times the string is going to be interpreted as it travels across various layers of the RPC interface.

PR #35, opened 2017-01-06
#37239: Use the new tree widget for Loader hierarchy navigation.

fpainchaud:
Handle a new configuration setting option that adds the capability to define left hand side tabs driven by the new Shotgun hierarchy model.

### DIFF
--- a/python/tk_multi_loader/delegate_publish_list.py
+++ b/python/tk_multi_loader/delegate_publish_list.py
@@ -19,6 +19,7 @@ shotgun_view = sgtk.platform.import_framework("tk-framework-qtwidgets", "views")
 
 from .ui.widget_publish_list import Ui_PublishListWidget
 from .delegate_publish import PublishWidget, PublishDelegate
+from . import model_item_data
 
 class PublishListWidget(PublishWidget):
     """
@@ -73,36 +74,12 @@ class SgPublishListDelegate(PublishDelegate):
         
         :param model_index: Model index to process
         :param widget: widget to adjust
-        """        
-        sg_data = shotgun_model.get_sg_data(model_index)
-        field_data = shotgun_model.get_sanitized_data(model_index, 
-                                                      shotgun_model.ShotgunModel.SG_ASSOCIATED_FIELD_ROLE)
+        """
 
-        # examples of data. In the examples below, when we say "node", this 
-        # means a node in the tree structure in the left hand side tree
-        
-        # intermediate node which isn't a link 
-        # field_data: {'name': 'sg_asset_type', 'value': 'Character' }
-        # sg_data:    None
-        
-        # intermediate node which is an entity
-        # field_data: {'name': 'sg_sequence', 'value': {'type': 'Sequence', 'id': 11, 'name': 'bunny_080'}}
-        # sg_data:    None
-        
-        # leaf node:  
-        # field_data: {'name': 'code', 'value': 'aaa_0030'}
-        # sg_data: {'code': 'aaa_00030',
-        #           'description': 'Created by the Shotgun Flame exporter.',
-        #           'id': 1662,
-        #           'image': 'https://....',
-        #           'project': {'id': 289, 'name': 'Climp', 'type': 'Project'},
-        #           'sg_sequence': {'id': 202, 'name': 'aaa', 'type': 'Sequence'},
-        #           'sg_status_list': 'wtg',
-        #           'type': 'Shot'}
-        
-        field_value = field_data["value"]
-        
-        # by default, just display the value 
+        # Extract the Shotgun data and field value from the model index.
+        (sg_data, field_value) = model_item_data.get_item_data(model_index)
+
+        # by default, just display the value
         main_text = field_value
         small_text = ""
 
@@ -115,7 +92,7 @@ class SgPublishListDelegate(PublishDelegate):
             # this can be a multi link field but also a field like a tags field or a non-entity link type field.
             formatted_values = []
             formatted_types = set()
-            
+
             for v in field_value:
                 if isinstance(v, dict) and "name" in v and "type" in v:
                     # This is a link field
@@ -125,7 +102,7 @@ class SgPublishListDelegate(PublishDelegate):
                         formatted_types.add(v["type"])
                 else:
                     formatted_values.append(str(v))
-            
+
             types = ", ".join(list(formatted_types))
             names = ", ".join(formatted_values)
             main_text = "<b>%s</b><br>%s" % (types, names)
--- a/python/tk_multi_loader/delegate_publish_thumb.py
+++ b/python/tk_multi_loader/delegate_publish_thumb.py
@@ -20,6 +20,7 @@ shotgun_view = sgtk.platform.import_framework("tk-framework-qtwidgets", "views")
 
 from .ui.widget_publish_thumb import Ui_PublishThumbWidget
 from .delegate_publish import PublishWidget, PublishDelegate
+from . import model_item_data
 
 class PublishThumbWidget(PublishWidget):
     """
@@ -78,22 +79,12 @@ class SgPublishThumbDelegate(PublishDelegate):
         :param model_index: Index of the item being drawn by the delegate.
         :param widget: Qt widget created by the delegate for rendering.
         """
-        # this is a publish!
-        sg_data = shotgun_model.get_sg_data(model_index)
+
+        # Extract the Shotgun data and field value from the model index.
+        (sg_data, field_value) = model_item_data.get_item_data(model_index)
 
         header_text = ""
         details_text = ""
-
-        # this is a folder item, injected into the publish model from the entity tree
-
-        field_data = shotgun_model.get_sanitized_data(model_index,
-                                                      shotgun_model.ShotgunModel.SG_ASSOCIATED_FIELD_ROLE)
-        # examples of data:
-        # intermediate node: {'name': 'sg_asset_type', 'value': 'Character' }
-        # intermediate node: {'name': 'sg_sequence',   'value': {'type': 'Sequence', 'id': 11, 'name': 'bunny_080'}}
-        # leaf node:         {'name': 'code',          'value': 'mystuff'}
-
-        field_value = field_data["value"]
 
         if isinstance(field_value, dict) and "name" in field_value and "type" in field_value:
             # intermediate node with entity link
--- a/python/tk_multi_loader/dialog.py
+++ b/python/tk_multi_loader/dialog.py
@@ -13,6 +13,7 @@ import sgtk
 from sgtk import TankError
 from sgtk.platform.qt import QtCore, QtGui
 
+from .model_hierarchy import SgHierarchyModel
 from .model_entity import SgEntityModel
 from .model_latestpublish import SgLatestPublishModel
 from .model_publishtype import SgPublishTypeModel
@@ -32,7 +33,6 @@ from .ui.dialog import Ui_Dialog
 
 # import frameworks
 shotgun_model = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_model")
-shotgun_fields = sgtk.platform.import_framework("tk-framework-qtwidgets", "shotgun_fields")
 settings = sgtk.platform.import_framework("tk-framework-shotgunutils", "settings")
 help_screen = sgtk.platform.import_framework("tk-framework-qtwidgets", "help_screen")
 overlay_widget = sgtk.platform.import_framework("tk-framework-qtwidgets", "overlay_widget")
@@ -997,7 +997,7 @@ class AppDialog(QtGui.QWidget):
 
             model = current_idx.model()
 
-            if not isinstance(model, (shotgun_model.SimpleShotgunHierarchyModel, SgEntityModel)):
+            if not isinstance(model, (SgHierarchyModel, SgEntityModel)):
                 # proxy model!
                 current_idx = model.mapToSource(current_idx)
 
@@ -1097,58 +1097,69 @@ class AppDialog(QtGui.QWidget):
         based on the config.
         """
         app = sgtk.platform.current_bundle()
-        entities = app.get_setting("entities")
 
-        for e in entities:
+        for setting_dict in app.get_setting("entities"):
 
-            # Validate that the settings dictionary contains all items needed.
+            # Validate that the setting dictionary contains all items needed.
+            # Here is an example of Hierarchy setting dictionary:
+            #     {'caption': 'Project',
+            #      'type':    'Hierarchy',
+            #      'root':    '/Project/{context.project.id}'}
+            # Here is an example of Query setting dictionary:
+            #     {'caption':     'My Tasks',
+            #      'type':        'Query',
+            #      'entity_type': 'Task',
+            #      'hierarchy':   ['project', 'entity', 'content'],
+            #      'filters':     [['task_assignees', 'is', '{context.user}'],
+            #                      ['project.Project.sg_status', 'is', 'Active']]}
 
-            k_error_msg = "Configuration error: One or more items in %s are missing a '%s' key!"
-            v_error_msg = "Configuration error: One or more items in %s have an invalid '%s' value for '%s' key!"
+            key_error_msg = "Configuration error: 'entities' item %s is missing key '%s'!"
+            value_error_msg = "Configuration error: 'entities' item %s key '%s' has an invalid value '%s'!"
 
-            k = "caption"
-            if k not in e:
-                raise TankError(k_error_msg % (entities, k))
+            key = "caption"
+            if key not in setting_dict:
+                raise TankError(key_error_msg % (setting_dict, key))
 
-            preset_name = e["caption"]
+            preset_name = setting_dict["caption"]
 
-            k = "type"
-            if k in e:
-                v = e[k]
-                if v not in ("Hierarchy", "Query"):
-                    raise TankError(v_error_msg % (entities, v, k))
-                model_type_hierarchy = (v == "Hierarchy")
+            key = "type"
+            if key in setting_dict:
+                value = setting_dict[key]
+                if value not in ("Hierarchy", "Query"):
+                    raise TankError(value_error_msg % (setting_dict, key, value))
+                type_hierarchy = (value == "Hierarchy")
             else:
-                model_type_hierarchy = False
+                # When the type is not given, default to "Query".
+                type_hierarchy = False
 
-            if model_type_hierarchy:
+            if type_hierarchy:
 
-                k = "root"
-                if k not in e:
-                    raise TankError(k_error_msg % (entities, k))
+                key = "root"
+                if key not in setting_dict:
+                    raise TankError(key_error_msg % (setting_dict, key))
 
                 sg_entity_type = "Project"
 
             else:
 
-                for k in ("entity_type", "hierarchy", "filters"):
-                    if k not in e:
-                        raise TankError(k_error_msg % (entities, k))
+                for key in ("entity_type", "hierarchy", "filters"):
+                    if key not in setting_dict:
+                        raise TankError(key_error_msg % (setting_dict, key))
 
-                sg_entity_type = e["entity_type"]
+                sg_entity_type = setting_dict["entity_type"]
 
             # get optional publish_filter setting
             # note: actual value in the yaml settings can be None, 
-            # that's why we cannot use e.get("publish_filters", []) 
-            publish_filters = e.get("publish_filters")
+            # that's why we cannot use setting_dict.get("publish_filters", [])
+            publish_filters = setting_dict.get("publish_filters")
             if publish_filters is None: 
                 publish_filters = []
 
             # Create the model.
-            if model_type_hierarchy:
-                (model, proxy_model) = self._setup_hierarchy_model(app, e["root"])
+            if type_hierarchy:
+                (model, proxy_model) = self._setup_hierarchy_model(app, setting_dict["root"])
             else:
-                (model, proxy_model) = self._setup_query_model(app, e)
+                (model, proxy_model) = self._setup_query_model(app, setting_dict)
 
             # Add a new tab and its layout to the main tab bar.
             tab = QtGui.QWidget()
@@ -1177,7 +1188,7 @@ class AppDialog(QtGui.QWidget):
                                           layout,
                                           view])
 
-            if not model_type_hierarchy:
+            if not type_hierarchy:
 
                 # Add a layout to host search.
                 search_layout = QtGui.QHBoxLayout()
@@ -1230,7 +1241,7 @@ class AppDialog(QtGui.QWidget):
             action_ca = QtGui.QAction("Collapse All Folders", view)
             action_ca.triggered.connect(view.collapseAll)
             view.addAction(action_ca)
-            if model_type_hierarchy:
+            if type_hierarchy:
                 action_refresh = QtGui.QAction("Reload", view)
                 action_refresh.triggered.connect(model.reload_data)
             else:
@@ -1287,7 +1298,7 @@ class AppDialog(QtGui.QWidget):
             root = None
 
         # Construct the hierarchy model.
-        model = shotgun_model.SimpleShotgunHierarchyModel(self, bg_task_manager=self._task_manager)
+        model = SgHierarchyModel(self, bg_task_manager=self._task_manager)
 
         # Create a proxy model.
         proxy_model = QtGui.QSortFilterProxyModel(self)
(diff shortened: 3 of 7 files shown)


python/tk_multi_loader/model_item_data.py
```diff
@@ -0,0 +1,88 @@
...
+
+shotgun_model = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_model")
+
+
+def get_item_data(item):
+    """
+    Extracts and standardizes the Shotgun data and field value from an item.
+
+    :param item: Selected item or model index.
+    :return: Standardized `(Shotgun data, field value)` extracted from the item data.
```
manneohrstrom: adding some examples here what 'standardized' means would be useful!

maybe those various examples that got deleted in other parts of the code could be shown here?

python/tk_multi_loader/model_item_data.py
```diff
@@ -0,0 +1,88 @@
...
+    #       "id", "name" and "type"
+    # - othewise, this value is the text data.
+    # - "value" can also be a list of such values.
+    # For an item in the ShotgunHierarchyModel tree structure,
+    # this data is None.
+    field_data = shotgun_model.get_sanitized_data(item, shotgun_model.ShotgunModel.SG_ASSOCIATED_FIELD_ROLE)
+
+    error_msg = "Unknown item '%s' model type!" % text_data
+
+    if field_data is None and \
```
manneohrstrom: is the idea that we try to introspect here and based on that try to figure out what kind of node it is?

Wouldn't it be better to just store the node type for each node as a custom role enum?

python/tk_multi_loader/model_latestpublish.py
```diff
@@ -309,22 +300,25 @@ def _load_external_data(self):
...
-            # copy across the std fields SG_ASSOCIATED_FIELD_ROLE and SG_DATA_ROLE
-            tree_item_sg_data = tree_view_item.get_sg_data()
-            item.setData(tree_item_sg_data, SgLatestPublishModel.SG_DATA_ROLE)
+            # Extract the Shotgun data and field value from the tree view item.
+            (tree_view_sg_data, field_value) = model_item_data.get_item_data(tree_view_item)
 
-            tree_item_field_data = tree_view_item.data(shotgun_model.ShotgunModel.SG_ASSOCIATED_FIELD_ROLE)
-            item.setData(tree_item_field_data, SgLatestPublishModel.SG_ASSOCIATED_FIELD_ROLE)
+            # Rebuild tree view item field data.
+            # Since we do not care about the "name" value, arbitrarily use "data".
```
manneohrstrom: would it be easier to read this code if there was a big 

```
if tree_view_item comes from a hierarchy model do X else do Y?
```

The `# Since we do not care about the "name" value, arbitrarily use "data".` confuses me. What does it mean? If we don't care about it, why is it there?

python/tk_multi_loader/model_item_data.py
```diff
@@ -0,0 +1,88 @@
...
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sgtk
+from sgtk import TankError
+from sgtk.platform.qt import QtCore
+
+shotgun_model = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_model")
+
+
+def get_item_data(item):
```
manneohrstrom: are we compensating for some methods that are missing on the sg item classes themselves? could this be done by implementing a `get_item_data()` on each item type instead? Sorry if this is a dumb idea, just trying to think of ways to make it a little bit clearer.

python/tk_multi_loader/delegate_publish_list.py
```diff
@@ -73,36 +74,12 @@ def _format_folder(self, model_index, widget):
         
         :param model_index: Model index to process
         :param widget: widget to adjust
-        """        
-        sg_data = shotgun_model.get_sg_data(model_index)
-        field_data = shotgun_model.get_sanitized_data(model_index, 
-                                                      shotgun_model.ShotgunModel.SG_ASSOCIATED_FIELD_ROLE)
+        """
 
-        # examples of data. In the examples below, when we say "node", this
```
fpainchaud: Commenting the details of `sg_data` and `field_data` was moved to `model_item_data.get_item_data(item)` in order to maintain these comments in one location and avoid the current discrepancies here and there.

python/tk_multi_loader/delegate_publish_thumb.py
```diff
@@ -78,23 +79,13 @@ def _format_folder(self, model_index, widget):
...
+        (sg_data, field_value) = model_item_data.get_item_data(model_index)
 
         header_text = ""
         details_text = ""
 
-        # this is a folder item, injected into the publish model from the entity tree
-
-        field_data = shotgun_model.get_sanitized_data(model_index,
-                                                      shotgun_model.ShotgunModel.SG_ASSOCIATED_FIELD_ROLE)
-        # examples of data:
```
fpainchaud: Commenting the details of `sg_data` and `field_data` was moved to `model_item_data.get_item_data(item)` in order to maintain these comments in one location and avoid the current discrepancies here and there.

python/tk_multi_loader/dialog.py
```diff
@@ -1265,6 +1269,110 @@ def _load_entity_presets(self):
...
+        Create the model and proxy model required by a hierarchy type configuration setting.
+
+        :param app: :class:`Application`, :class:`Engine` or :class:`Framework` bundle instance
+                    associated with the loader.
+        :param root: The path to the root of the Shotgun hierarchy to display.
+        :return: Created `(model, proxy model)`.
+        """
+
+        # Resolve any magic tokens in the root.
+        if app.context.project:
```
fpainchaud: These are two very different configuration settings:
- The "root" one is simple and involves replacing the "magic token" in the root path.
- The "filters" one is complex and involves replacing the "magic tokens" in a list of of lists.
Centralizing into a single method would not bring any benefit and would make the code harder to understand and maintain.

python/tk_multi_loader/dialog.py
```diff
@@ -1265,6 +1269,110 @@ def _load_entity_presets(self):
...
+
+        # Create a proxy model.
+        proxy_model = QtGui.QSortFilterProxyModel(self)
+        proxy_model.setSourceModel(model)
+
+        # Impose and keep the sorting order on the default display role text.
+        proxy_model.sort(0)
+        proxy_model.setDynamicSortFilter(True)
+
+        # The field manager handles retrieving widgets for Shotgun field types.
```
fpainchaud: This code is based on Josh's demo code. @josh-t ?

python/tk_multi_loader/dialog.py
```diff
@@ -1265,6 +1269,110 @@ def _load_entity_presets(self):
...
+        # It needs time to initialize itself, after which the widgets can begin to be populated.
+        # "PublishedFile.entity" means build a hierarchy that leads to
+        # entities that are linked via the PublishedFile.entity field.
+        entity_fields = {"Asset": [ # Fields usually added by the configuration setting.
```
fpainchaud: Without this `entity_fields` dictionary, the data loaded by the hierarchy model would be missing some item data needed by `model_item_data.get_item_data(item)` for _Asset_ and _Shot_ leaf items.

Is there any other entries I should add in this dictionary?

python/tk_multi_loader/dialog.py
```diff
@@ -1322,7 +1430,8 @@ def _switch_profile_tab(self, new_index, combo_operation_mode):
             # history navigation, ask the currently visible
             # view to (background async) refresh its data
             model = self._entity_presets[self._current_entity_preset].model
-            model.async_refresh()
+            if isinstance(model, SgEntityModel):
+                model.async_refresh()
```
fpainchaud: I tried this approach of keeping the tab "type", but this proved less intuitive since the connection between a tab "type" and its model type had to be kept in mind and extra commented, not to mention harder to maintain if/when a model type changes.

I will add a comment to the current code.

python/tk_multi_loader/model_item_data.py
fpainchaud: All the Loader code in based on (expects) certain `SG_DATA_ROLE` and `SG_ASSOCIATED_FIELD_ROLE` data formats retrieved from `ShotgunModel` items.

For `ShotgunHierarchyModel` items, `SG_DATA_ROLE` data has a very different format, and there is no `SG_ASSOCIATED_FIELD_ROLE` data.

Centralizing the _standardization_ of the item data in this function was the cleaner approach with the current `ShotgunModel` and `ShotgunHierarchyModel` item data.

python/tk_multi_loader/model_item_data.py
fpainchaud: We try to introspect to decide if the item or model index is coming from a `ShotgunModel` or a `ShotgunHierarchyModel`. Being able to query a custom "model type" role to know this would be great.

python/tk_multi_loader/dialog.py
josh-t: What about studios that use different entity types for their hierarchy? What does the loader do with the fields that aren't there by default? Is it displaying them on selection somehow?

python/tk_multi_loader/dialog.py
```diff
@@ -1265,6 +1269,110 @@ def _load_entity_presets(self):
...
+
+        return (model, proxy_model)
+
+    def _setup_query_model(self, app, e):
+        """
+        Create the model and proxy model required by a query type configuration setting.
+
+        :param app: :class:`Application`, :class:`Engine` or :class:`Framework` bundle instance
+                    associated with the loader.
+        :param e: Configuration setting dictionary for a tab.
```
josh-t: Can we change the name of this variable?

python/tk_multi_loader/dialog.py
```diff
@@ -995,7 +997,7 @@ def _get_selected_entity(self):
 
             model = current_idx.model()
 
-            if not isinstance(model, SgEntityModel):
+            if not isinstance(model, (shotgun_model.SimpleShotgunHierarchyModel, SgEntityModel)):
```
fpainchaud: Where should this be done?

python/tk_multi_loader/dialog.py
```diff
@@ -1099,11 +1101,41 @@ def _load_entity_presets(self):
 
         for e in entities:
 
-            # validate that the settings dict contains all items needed.
-            for k in ["caption", "entity_type", "hierarchy", "filters"]:
+            # Validate that the settings dictionary contains all items needed.
```
fpainchaud: Done.

python/tk_multi_loader/dialog.py
```diff
@@ -1099,11 +1101,41 @@ def _load_entity_presets(self):
...
+
+            k_error_msg = "Configuration error: One or more items in %s are missing a '%s' key!"
+            v_error_msg = "Configuration error: One or more items in %s have an invalid '%s' value for '%s' key!"
+
+            k = "caption"
```
fpainchaud: Done.

python/tk_multi_loader/dialog.py
fpainchaud: Done.

python/tk_multi_loader/dialog.py
```diff
@@ -1099,11 +1101,41 @@ def _load_entity_presets(self):
...
+            if k not in e:
+                raise TankError(k_error_msg % (entities, k))
+
+            preset_name = e["caption"]
```
fpainchaud: It would just mean duplicating the `'caption'` validation and would not really create two fully separate branches since the `'type'` has to be tested on its own before branching. Using better variable names already made the code easier to read.

python/tk_multi_loader/dialog.py
```diff
@@ -1112,147 +1144,119 @@ def _load_entity_presets(self):
...
-            action_refresh.triggered.connect(model.async_refresh)
             view.addAction(action_ea)
+            action_ca = QtGui.QAction("Collapse All Folders", view)
+            action_ca.triggered.connect(view.collapseAll)
             view.addAction(action_ca)
+            if model_type_hierarchy:
+                action_refresh = QtGui.QAction("Reload", view)
+                action_refresh.triggered.connect(model.reload_data)
+            else:
+                action_refresh = QtGui.QAction("Refresh", view)
```
fpainchaud: The action menu implements a `Reload` for hierarchy tabs. This is implemented with a `reload_data()` convenience method in a new class derived from `SimpleShotgunHierarchyModel` living in the Loader.

python/tk_multi_loader/dialog.py
fpainchaud: Unneeded field manager removed.

python/tk_multi_loader/model_latestpublish.py
fpainchaud: Indeed, no need for key "name". I updated this line of code and its comment.

python/tk_multi_loader/dialog.py
fpainchaud: After investigation to understand what is possible, I updated the code and commented:

`# We need to provide a dictionary that identifies what additional fields to include`
`# for the loaded hierarchy leaf entities in addition to "id" and "type".`
`# When available, these fields will be used to add info in the detail panel.`

Ideally, the entity type list for entities with publishes should be retrieved from the settings.

python/tk_multi_loader/dialog.py
fpainchaud: Current code commented.

python/tk_multi_loader/model_item_data.py
fpainchaud: It would certainly be a lot cleaner to make the items return one known format of data, whatever the model class type they belong to.

For the time being, I have added purpose details in the function doc string.

python/tk_multi_loader/model_item_data.py
fpainchaud: Added detailed examples for various sources of items.

fpainchaud:
Gentlemen, let's go for a second CR.

python/tk_multi_loader/dialog.py
```diff
@@ -1095,164 +1097,177 @@ def _load_entity_presets(self):
...
+            # Set up a view right click menu.
             action_ea = QtGui.QAction("Expand All Folders", view)
-            action_ca = QtGui.QAction("Collapse All Folders", view)
-            action_refresh = QtGui.QAction("Refresh", view)
-
             action_ea.triggered.connect(view.expandAll)
-            action_ca.triggered.connect(view.collapseAll)
-            action_refresh.triggered.connect(model.async_refresh)
             view.addAction(action_ea)
+            action_ca = QtGui.QAction("Collapse All Folders", view)
```
manneohrstrom: seems like expand all folders shouldn't be allowed for the hierarchy model - can we take it out?

python/tk_multi_loader/dialog.py
```diff
@@ -1265,6 +1280,94 @@ def _load_entity_presets(self):
...
+
+        # Resolve any magic tokens in the root.
+        if app.context.project:
+            project_id = app.context.project.get("id")
+            root = root.replace("{context.project.id}", str(project_id))
+        else:
+            root = None
+
+        # Construct the hierarchy model.
+        model = SgHierarchyModel(self, bg_task_manager=self._task_manager)
```
manneohrstrom: i think i'd prefer if we would refactor this into a class, like it's done for the sg model: https://github.com/shotgunsoftware/tk-multi-loader2/blob/master/python/tk_multi_loader/model_entity.py#L54

python/tk_multi_loader/dialog.py
```diff
@@ -1265,6 +1280,94 @@ def _load_entity_presets(self):
...
+        proxy_model.setSourceModel(model)
+
+        # Impose and keep the sorting order on the default display role text.
+        proxy_model.sort(0)
+        proxy_model.setDynamicSortFilter(True)
+
+        # We need to provide a dictionary that identifies what additional fields to include
+        # for the loaded hierarchy leaf entities in addition to "id" and "type".
+        # When available, these fields will be used to add info in the detail panel.
+        # TODO: Our entity type list for entities with publishes should be retrieved from the settings.
```
manneohrstrom: We have this sort of thing in [core](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/context.py#L1441):

```python
def _get_entity_type_sg_name_field(entity_type):
    """
    Return the Shotgun name field to use for the specified entity type.  This
    is needed as not all entity types are consistent!

    :param entity_type:     The entity type to get the name field for
    :returns:               The name field for the specified entity type
    """
    return {"HumanUser":"name", 
            "Task":"content", 
            "Project":"name"}.get(entity_type, "code")
```

I think we should adopt a similar approach. For description and sg_status_list, we can just gracefully fall back  and handle the case when the field is not returned for something. Do you think that could work?

python/tk_multi_loader/dialog.py
```diff
@@ -1265,6 +1280,94 @@ def _load_entity_presets(self):
...
+        entity_fields = {}
+        for entity_type in ["Asset", "Shot"]:
```
josh-t: I'm wondering if it would be possible (and in if so, how slow would it be) to use the python api's `summarize` method to figure out the various types of entities that `PublishedFile`s on the project/site are linked to. I haven't used `summarize` much at all, so this could be a silly comment. But if that were possible, then you could use the schema to check those entity type's fields and make some educated guesses (with fallbacks) as to what fields should be returned for each type.

python/tk_multi_loader/model_item_data.py
```diff
@@ -0,0 +1,188 @@
...
+    # Example for a ShotgunHierarchyModel intermediate item (without any extra entity fields requested):
+    #     {'has_children':    True,
+    #      'label':           'Character',
+    #      'path':            '/Project/70/Asset/sg_asset_type/Character',
+    #      'ref':             {'kind': 'list', 'value': 'Character'},
+    #      'target_entities': {'additional_filter_presets': [{'path':        '/Project/70/Asset/sg_asset_type/Character',
+    #                                                         'preset_name': 'NAV_ENTRIES',
+    #                                                         'seed':        {'field': 'entity', 'type': 'PublishedFile'}}],
+    #                          'type': 'PublishedFile'}}
+    #
```
josh-t: For readability, I'd suggest using a form like this: 

```
{
    'has_children': True,
    'label': 'Character',
    'path': '/Project/70/Asset/sg_asset_type/Character',
    'ref': {
        'kind': 'list', 
        'value': 'Character'
    },
    'target_entities': {
        'additional_filter_presets': [
            {
                'path': '/Project/70/Asset/sg_asset_type/Character',
                'preset_name': 'NAV_ENTRIES',
                'seed': {
                    'field': 'entity',
                    'type': 'PublishedFile'
                }
            }
        ],
        'type': 'PublishedFile'
    }
}
```

python/tk_multi_loader/model_item_data.py
```diff
@@ -0,0 +1,188 @@
...
+    #      'ref':             {'kind': 'entity', 'value': {'id':             1230,
+    #                                                      'type':           'Asset',
+    #                                                      'code':           'Bunny',
+    #                                                      'description':    '...',
+    #                                                      'image':          '/thumbnail/Asset/1230?567',
+    #                                                      'sg_status_list': 'fin'}},
+    #      'target_entities': {'additional_filter_presets': [{'path':        '/Project/70/Asset/sg_asset_type/Character/id/1230',
+    #                                                         'preset_name': 'NAV_ENTRIES',
+    #                                                         'seed':        {'field': 'entity', 'type': 'PublishedFile'}}],
+    #                          'type': 'PublishedFile'}}
```
josh-t: same comment about readability.

python/tk_multi_loader/model_item_data.py
```diff
@@ -0,0 +1,188 @@
...
+            # We have an intermediate item.
+            # For example:
+            #     {'has_children':    True,
+            #      'label':           'Character',
+            #      'path':            '/Project/70/Asset/sg_asset_type/Character',
+            #      'ref':             {'kind': 'list', 'value': 'Character'},
+            #      'target_entities': {'additional_filter_presets': [{'path':        '/Project/70/Asset/sg_asset_type/Character',
+            #                                                         'preset_name': 'NAV_ENTRIES',
+            #                                                         'seed':        {'field': 'entity', 'type': 'PublishedFile'}}],
+            #                          'type': 'PublishedFile'}}
```
josh-t: same comment about readability.

python/tk_multi_loader/model_item_data.py
```diff
@@ -0,0 +1,188 @@
...
+            #      'ref':             {'kind': 'entity', 'value': {'id':             1230,
+            #                                                      'type':           'Asset',
+            #                                                      'code':           'Bunny',
+            #                                                      'description':    '...',
+            #                                                      'image':          '/thumbnail/Asset/1230?567',
+            #                                                      'sg_status_list': 'fin'}},
+            #      'target_entities': {'additional_filter_presets': [{'path':        '/Project/70/Asset/sg_asset_type/Character/id/1230',
+            #                                                         'preset_name': 'NAV_ENTRIES',
+            #                                                         'seed':        {'field': 'entity', 'type': 'PublishedFile'}}],
+            #                          'type': 'PublishedFile'}}
```
josh-t: same comment about readability.

python/tk_multi_loader/model_item_data.py
```diff
@@ -0,0 +1,188 @@
...
+    #     {'name': 'code', 'value': 'Bunny'}
+    #     {'name': 'code', 'value': 'bunny_010_0010'}
+    #
+    # For an item in the ShotgunHierarchyModel tree structure, this data is None.
+    #
+    field_data = shotgun_model.get_sanitized_data(item, shotgun_model.ShotgunModel.SG_ASSOCIATED_FIELD_ROLE)
+
+    if field_data is None and \
+       isinstance(sg_data, dict) and "has_children" in sg_data and "ref" in sg_data and \
+       isinstance(sg_data["ref"], dict) and "value" in sg_data["ref"]:
```
josh-t: To avoid hardcoding this knowledge of the dict structure in the conditional, could you check the model type via `isinstance(item.index().model(), ...)`?

python/tk_multi_loader/model_item_data.py
josh-t: If you haven't already, want to create a ticket to look into adding additional methods on the sg model and hierarchy model item classes to help out here? maybe make a reference to that future work here? Sounds like we're all in agreement about how to make this better, just unsure of the implementation at this point?

python/tk_multi_loader/model_item_data.py
fpainchaud: I had tried this approach initially, but it was not that straightforward considering the various types of model and index items that can be passed in. Though, I just revisited this approach following your suggestion and was able to circumscribe the cases that need to be tested.

python/tk_multi_loader/dialog.py
```diff
@@ -1095,164 +1097,177 @@ def _load_entity_presets(self):
...
+            view.setHeaderHidden(True)
+            view.setModel(proxy_model)
 
-            # set up right click menu
+            # Keep a handle to all the new Qt objects, otherwise the GC may not work.
+            self._dynamic_widgets.extend([model,
+                                          proxy_model,
+                                          tab,
+                                          layout,
+                                          view])
```
fpainchaud: Done.

python/tk_multi_loader/model_item_data.py
fpainchaud: Done.

python/tk_multi_loader/model_item_data.py
fpainchaud: Done.

python/tk_multi_loader/model_item_data.py
fpainchaud: Done.

python/tk_multi_loader/model_item_data.py
fpainchaud: Done.

python/tk_multi_loader/model_item_data.py
fpainchaud: I will create a new ticket.

python/tk_multi_loader/dialog.py
fpainchaud: I will create a new ticket for this.

python/tk_multi_loader/dialog.py
fpainchaud: "Expand All Folders" gone for the hierarchy model.

python/tk_multi_loader/dialog.py
```diff
@@ -1095,164 +1097,177 @@ def _load_entity_presets(self):
...
-
             action_ea.triggered.connect(view.expandAll)
-            action_ca.triggered.connect(view.collapseAll)
-            action_refresh.triggered.connect(model.async_refresh)
             view.addAction(action_ea)
+            action_ca = QtGui.QAction("Collapse All Folders", view)
+            action_ca.triggered.connect(view.collapseAll)
             view.addAction(action_ca)
+            if type_hierarchy:
+                action_refresh = QtGui.QAction("Reload", view)
```
fpainchaud: Detailed tool tips added.

New code was needed to handle tool tip display for action context menus since this is not readily supported by Qt views.

python/tk_multi_loader/dialog.py
fpainchaud: Refactored into `SgHierarchyModel.__init__()`.

python/tk_multi_loader/dialog.py
fpainchaud: The issue is not to decide what fields are available for each entity, but what entities to put as keys in the `entity_fields` dictionary.

When a requested field is not available for a specific entity, it is simply ignored by the Shotgun model. This is already how the query model is working with such fields:
https://github.com/shotgunsoftware/tk-multi-loader2/blob/master/python/tk_multi_loader/model_entity.py#L54

The Loader already gracefully handles the lack of some fields when displaying the detail panel.

python/tk_multi_loader/dialog.py
```diff
@@ -1095,164 +1097,177 @@ def _load_entity_presets(self):
...
-        entities = app.get_setting("entities")
 
-        for e in entities:
+        for setting_dict in app.get_setting("entities"):
+
+            # Validate that the setting dictionary contains all items needed.
+            # Here is an example of Hierarchy setting dictionary:
+            #     {'caption': 'Project',
+            #      'type':    'Hierarchy',
+            #      'root':    '/Project/{context.project.id}'}
```
fpainchaud: `info.yml` updated.

PR #38, opened 2017-04-13
Add entity actions to multiloader2.

maxedoutstudio:
Added entity actions to multiloader2. Can now defined custom
'entity_actions' in the same manner as actions for publishes in the
config. These actions will then be subsequently added to the dropdown
for folders.

### DIFF
--- a/python/tk_multi_loader/loader_action_manager.py
+++ b/python/tk_multi_loader/loader_action_manager.py
@@ -81,11 +81,7 @@ class LoaderActionManager(ActionManager):
             raise TankError("Unsupported UI_AREA. Contact support.")
 
         # convert created_at unix time stamp to shotgun time stamp
-        unix_timestamp = sg_data.get("created_at")
-        if isinstance(unix_timestamp, float):
-            sg_timestamp = datetime.datetime.fromtimestamp(unix_timestamp,
-                                                           shotgun_api3.sg_timezone.LocalTimezone())
-            sg_data["created_at"] = sg_timestamp
+        self._fix_timestamp(sg_data)
 
         action_defs = []
         try:
@@ -273,7 +269,7 @@ class LoaderActionManager(ActionManager):
         :return: List of actions.
         """
 
-        publish_type = sg_data.get('type', None)
+        publish_type = sg_data.get("type", None)
 
         # check if we have logic configured to handle this publish type.
         mappings = self._app.get_setting("entity_mappings")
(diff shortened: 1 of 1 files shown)


python/tk_multi_loader/loader_action_manager.py
```diff
@@ -265,7 +264,47 @@ def has_actions(self, publish_type):
...
+
+    def _get_actions_for_folder(self, sg_data):
+        """
+        Retrieves the list of actions for a given folder.
+
+        :param sg_data: Publish to retrieve actions for
+        :return: List of actions.
+        """
+
+        publish_type = sg_data.get('type', None)
```
jfboismenu: Please always use double quotes.

python/tk_multi_loader/loader_action_manager.py
```diff
@@ -265,7 +264,47 @@ def has_actions(self, publish_type):
...
+
+        # returns a structure on the form
+        # { "Shot": ["reference", "import"] }
+        actions = mappings.get(publish_type, [])
+
+        if len(actions) == 0:
+            return []
+
+        # convert created_at unix time stamp to shotgun time stamp
+        unix_timestamp = sg_data.get("created_at")
```
jfboismenu: Could you refactor this timestamp fix bit into its own method and update the other place in the file that uses this? Thanks!

python/tk_multi_loader/loader_action_manager.py
```diff
@@ -276,26 +315,60 @@ def get_actions_for_folder(self, sg_data):
         :return: List of QAction instances.
         """
 
-        actions = []
+        qt_actions = []
+
+        # If the selection is empty, we skip this and append default actions
```
jfboismenu: I think your comment might be out of date. It looks like if the selection is not empty, we add some custom actions and then the default actions.

python/tk_multi_loader/loader_action_manager.py
```diff
@@ -276,26 +315,60 @@ def get_actions_for_folder(self, sg_data):
...
+        if len(sg_data) != 0:
+
+            # Gets the actions for the folder
+            entity_actions = self._get_actions_for_folder(sg_data)
+
+            # For every actions in the intersection, create an associated QAction with appropriate callback
```
jfboismenu: ~in the intersection~

jfboismenu:
@manneohrstrom What do you think, should be add multi-selection support for folders as well, which means we would only show the intersection of actions for the folders, just like we do for publishes. If this is not required for the Flame workflows, maybe we can ticket it up as an enhancement for the future?

manneohrstrom:
@jfboismenu that sounds like a great idea! Seems nice with consistency and yes i think ticketing it up sounds ideal if there isn't an immediate need.

maxedoutstudio:
@manneohrstrom Confirmed from Flame side that it isn't an immediate requirement for us. Let me know whenever we can merge this.

PR #39, opened 2017-04-13
Supports auto-extraction of format/frame range for zero config.

josh-t:
This change is specifically for movie formats. I think I'll have to come back to it for image sequences and make tweaks.

### DIFF
--- a/hooks/tk-nuke_actions.py
+++ b/hooks/tk-nuke_actions.py
@@ -183,11 +183,11 @@ class NukeActions(HookBaseClass):
         if ext.lower() not in valid_extensions:
             raise Exception("Unsupported file extension for '%s'!" % path)
 
-        # This form will extract the format and frame range from the file
-        # itself. We can also see if there's a matching template and override
-        # the frame range, but this should handle the zero config case. This
-        # will also automatically extract the format and frame range for movie
-        # files.
+        # `nuke.createNode()` will extract the format and frame range from the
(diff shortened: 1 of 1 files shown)


hooks/tk-nuke_actions.py
```diff
@@ -183,15 +183,21 @@ def _create_read_node(self, path, sg_publish_data):
         if ext.lower() not in valid_extensions:
             raise Exception("Unsupported file extension for '%s'!" % path)
 
+        # This form will extract the format and frame range from the file
+        # itself. We can also see if there's a matching template and override
+        # the frame range, but this should handle the zero config case. This
+        # will also automatically extract the format and frame range for movie
+        # files.
+        read_node = nuke.createNode("Read")
```
jfboismenu: Hi Josh, the code looks alright, but I have no idea what is going on. To me it looks like you’re just instantiating a node and setting its values one by one instead of the old code which instantiated Node with all values already set.

Why the change?

jfboismenu: If there's a technical reason for doing it this way instead of the other, which to an untrained eye looks the same, maybe its worth pointing it out in the code. Otherwise you run the risk of someone seeing that code and think "Hey, why is this not simply using `nuke.nodes.ReadNode`, it would be a single line of code instead of three."

josh-t:
manually merged.

PR #40, opened 2017-04-18
For #40813, hierarchical refresh issue

jfboismenu:
When clicking on a node in the left-hand side, we `fetchMore`, which asynchronously updates the model. The issue is that we don't listen for any events about the model being updated after this has been called. Because of this, any folder items that should be displayed are not.

This fixes the issue.

### DIFF
--- a/python/tk_multi_loader/dialog.py
+++ b/python/tk_multi_loader/dialog.py
@@ -1345,7 +1345,22 @@ class AppDialog(QtGui.QWidget):
         proxy_model.sort(0)
         proxy_model.setDynamicSortFilter(True)
 
+        # When clicking on a node, we fetch all the nodes under it so we can populate the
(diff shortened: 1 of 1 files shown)


python/tk_multi_loader/dialog.py
```diff
@@ -1345,8 +1345,23 @@ def _setup_hierarchy_model(self, app, root):
...
+        # When clicking on a node, we fetch all the nodes under it so we can populate the
+        # right hand-side. Make sure we are notified when the child come back so we can load
+        # publishes for the current item.
+        model.data_refreshed.connect(self._hierarchy_refreshed)
+
         return (model, proxy_model)
 
+    def _hierarchy_refreshed(self):
+        """
+        Signal triggered when the hierarchy model has been refreshed. This allows to show all the
```
josh-t: suggest: `Slot triggered when...`

PR #45, opened 2017-05-03
#41668 Adds top-level sibling to display project publishes.

josh-t:
This change uses the `include_root` option to the sg hierarchy model to display a top-level project item in the hierarchy. This makes it possible to load project-level publishes. 

screen shot 2017-05-02 at 11 47 39 pm


python/tk_multi_loader/dialog.py
```diff
@@ -1418,9 +1418,20 @@ def _setup_hierarchy_model(self, app, root):
         :return: Created `(model, proxy model)`.
         """
 
+        # If the root is a aproject, include it in the hierarchy model so that
+        # we can display project publishes. We do an innocent little hack here
+        # by including a space at the front of the project root item to make it
+        # display first in the tree.
+        include_root = " Project Publishes" if root.get("type") == "Project" else None
```
jfboismenu: I would prefer if you didn't do this hack and instead derived a proxy model from QSortFilterproxyModel and reimplement [lessThan](http://doc.qt.io/qt-4.8/qsortfilterproxymodel.html#lessThan).

python/tk_multi_loader/dialog.py
```diff
@@ -1418,9 +1418,20 @@ def _setup_hierarchy_model(self, app, root):
         :return: Created `(model, proxy model)`.
         """
 
+        # If the root is a aproject, include it in the hierarchy model so that
```
jfboismenu: ~a~project

python/tk_multi_loader/dialog.py
jfboismenu: I know its more complicated, but I think it would be great if someone on the team got to learn how this works so we can do cooler things with proxy models in the future.

Also, maybe its because I know, but the spacing between the icon and the text shows. :p

josh-t:
Manually merged a while back.

PR #47, opened 2017-05-04
Fix disappearing folder when clicking them

mtiudaeu:
The bug was reproducible when the entity_mappings was not defined in the config.

### DIFF
--- a/info.yml
+++ b/info.yml
@@ -73,6 +73,12 @@ configuration:
         default_value_tk-mari:
             Alembic Cache: [geometry_import]
 
+    entity_mappings:
+        type: dict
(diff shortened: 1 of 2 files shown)


info.yml
```diff
@@ -73,6 +73,12 @@ configuration:
         default_value_tk-mari:
             Alembic Cache: [geometry_import]
 
+    entity_mappings:
+        type: dict
+        description: Associates entity types with actions. The actions are all defined
+                     inside the actions hook, so a number of actions are available by default
```
jfboismenu: Could you remove `, so a number of actions are available by default and in addition to this you can add your own actions if you like` (don't forget to add a period at the end :p)? There are no defaults.

PR #48, opened 2017-05-26
For #39658, loader notifications when action is launched.

jfboismenu:
This adds a banner that is displayed before running a user action. The banner will be dismissed as early as after 3 seconds it was displayed, but not before control the hook has finished executing.

### DIFF
--- a/python/tk_multi_loader/banner.py
+++ b/python/tk_multi_loader/banner.py
@@ -14,12 +14,23 @@ import time
 
 
 class Banner(QtGui.QLabel):
+    """
+    Banner that can be shown and then dismissed with an animation. Will always
+    be shown for at least 3 seconds even if a request to hide it is done before
+    the time is up. The banner will always be displayed at the top of the parent
+    widget's window.
+    """
 
+    # Height of the widget.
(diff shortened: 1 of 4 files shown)


python/tk_multi_loader/banner.py
```diff
@@ -0,0 +1,111 @@
...
+    _HEIGHT = 32
+
+    def __init__(self, parent):
+        """
+        :param parent: Parent widget.
+        """
+        super(Banner, self).__init__(parent)
+
+        # Sets the style sheet for the widget.
+        self.setStyleSheet("""
```
josh-t: This should be defined here I think: https://github.com/shotgunsoftware/tk-multi-loader2/blob/master/style.qss

Any reason not to?

jfboismenu: The style sheet applies to the area under the builtin info banner at the top of everything Toolkit app. However, the banner is a child of the dialog itself, so the CSS doesn't apply.

jfboismenu:
This was properly merged into master. Don't know what Github is talking about...

PR #50, opened 2017-06-12
SMOK-45222 | tk-flame_actions.py implementation

krbg:
This implementation adds a tk-flame-action implementation to the loader.

The actions are: 

Import clip: Load a clip into flame into "Schematic Reel 1" of the current batch setup.

Load Batch Group: Load a batch setup to the current Batch Group.

Create a Batch Group: Load the selected Shot in flame if a .batch file is present in a new Batch Group.

### DIFF
--- a/hooks/tk-flame_actions.py
+++ b/hooks/tk-flame_actions.py
@@ -234,16 +234,22 @@ class FlameActions(HookBaseClass):
         :type sg_publish_data: dict
         """
         # Determines published files
-        sg_filters = [["entity", "is", sg_publish_data]]
-        sg_fields = ["path", "published_file_type", "version"]
-        sg_type = "PublishedFile"
+        sg_filters = [["id", "is", sg_publish_data["id"]]]
+        sg_fields = ["sg_published_files",
+                     "code"
+                     ]
 
-        published_files = self.parent.shotgun.find(
+        sg_type = sg_publish_data["type"]
+
+        sg_info = self.parent.shotgun.find_one(
             sg_type, filters=sg_filters, fields=sg_fields
         )
 
-        batch_path = self._get_batch_path_from_published_files(published_files)
+        # Checks that we have the necessary info to proceed.
+        if not all(f in sg_info for f in sg_fields):
+            raise FlameActionError("Cannot load a Batch Group from Shotgun using this {}".format(sg_type))
 
+        batch_path = self._get_batch_path_from_published_files(sg_info)
         if batch_path and os.path.exists(batch_path):
             flame.batch.create_batch_group(sg_publish_data["code"])
             flame.batch.go_to()
@@ -308,13 +314,20 @@ class FlameActions(HookBaseClass):
         published_files = []
 
         for published_file in sg_published_files:
-            path = self.get_publish_path(published_file)
+            # Gets paths to published files
+            sg_filters = [["id", "is", published_file["id"]]]
+            sg_fields = ["path", "published_file_type", "version"]
+            sg_type = "PublishedFile"
+
+            file_info = self.parent.shotgun.find_one(sg_type, filters=sg_filters, fields=sg_fields)
+
+            path = self.get_publish_path(file_info)
 
             # Eliminates PublishedFiles with an invalid local path
             if path and os.path.exists(path):
-                published_files.append({"path": path, "info": published_file})
+                published_files.append({"path": path, "info": file_info})
             elif '%' in path:
-                sg_filters = [["id", "is", published_file["version"]["id"]]]
+                sg_filters = [["id", "is", file_info["version"]["id"]]]
                 sg_fields = ["frame_range"]
                 sg_type = "Version"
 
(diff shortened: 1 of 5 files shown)


jpbrault:
Are all these commit expected?

hooks/tk-flame_actions.py
```diff
@@ -0,0 +1,383 @@
...
+        """
+        app = self.parent
+        app.log_debug("Execute action called for action %s. "
+                      "Parameters: %s. Publish Data: %s" % (name, params, sg_publish_data))
+
+        try:
+            if name == "load_clip":
+                self._import_clip(sg_publish_data)
+
+            if name == "load_setup":
```
jpbrault: Could 'elif'

hooks/tk-flame_actions.py
```diff
@@ -0,0 +1,383 @@
...
+                self._import_batch_file(sg_publish_data)
+
+            if name == "load_batch":
+                self._import_batch_group(sg_publish_data)
+        except FlameActionError as error:
```
jpbrault: else: Error?

krbg:
@jpbrault I took back an old branch so yes..

hooks/tk-flame_actions.py
```diff
@@ -0,0 +1,383 @@
...
+
+        sg_info = self.parent.shotgun.find_one(
+            sg_type, filters=sg_filters, fields=sg_fields
+        )
+
+        # Checks that we have the necessary info to proceed.
+        if not all(f in sg_info for f in sg_fields):
+            raise FlameActionError("Cannot load a Batch Group from Shotgun using this {}".format(sg_type))
+
+        published_files = self._get_paths_from_published_files(
```
mtiudaeu: [minor] The naming here is a bit confusing. published_files is published_file_paths. Also _get_batch_path_from_published_files could call _get_paths_from_published_files directly.

hooks/tk-flame_actions.py
```diff
@@ -0,0 +1,383 @@
...
+                self._import_clip(sg_publish_data)
+
+            if name == "load_setup":
+                self._import_batch_file(sg_publish_data)
+
+            if name == "load_batch":
+                self._import_batch_group(sg_publish_data)
+        except FlameActionError as error:
+            from PySide import QtGui, QtCore
+            if QtCore.QCoreApplication.instance():
```
jfboismenu: Could you elaborate why this check is required? Why would the Flame loader hook be run without a QApplication ready? At the very least you should check for  QtCore.QApplication.instance(), since you want to show a UI popup.

hooks/tk-flame_actions.py
jfboismenu: The methods being called all raise an exception. This is correct.

hooks/tk-flame_actions.py
```diff
@@ -0,0 +1,383 @@
...
+                QtGui.QMessageBox.information(None, "Flame Action Error", str(error))
```
jfboismenu: The [loader app](https://github.com/shotgunsoftware/tk-multi-loader2/blob/master/python/tk_multi_loader/loader_action_manager.py#L397-L407) already catches errors. Why is this required?

jfboismenu: I would be annoyed if I imported 10 things and they all failed and got 10 pop-ups.

jfboismenu: Maybe logging a warning would be better? How would that show up in the Flame UI?

hooks/tk-flame_actions.py
```diff
@@ -0,0 +1,383 @@
...
+    def _import_batch_group(self, sg_publish_data):
+        """
+        Imports a Batch Group from Shotgun into Flame.
+
+        :param sg_publish_data: Shotgun data dictionary with all the standard publish fields.
+        :type sg_publish_data: dict
+        """
+        # Determines published files
+        sg_filters = [["id", "is", sg_publish_data["id"]]]
+        sg_fields = ["sg_published_files",
```
jfboismenu: I'm worried here. The `sg_` are not part of the `PublishedFile` entity in Shotgun. Are those fields you have added on your site? Clients will not have these fields and running this hook will fail. We should grab some time offline to discuss this. Please ping me on Slack so we can schedule something.

python/tk_multi_loader/delegate_publish_list.py
```diff
@@ -165,7 +165,11 @@ def _format_publish(self, model_index, widget):
         # Publish Name Version 002
         sg_data = shotgun_model.get_sg_data(model_index)
         main_text = "<b>%s</b>" % (sg_data.get("name") or "Unnamed")
-        main_text += " Version %03d" % sg_data.get("version_number")        
+
+        version = sg_data.get("version_number")
+        vers_str = "%03d" % version if version else "N/A"
```
jfboismenu: `if version is not None` would be better. It is possible to publish version 0 (I've seen it before!)

python/tk_multi_loader/dialog.py
```diff
@@ -989,9 +993,11 @@ def _pre_execute_action(self, action):
             sg_data = data[0]["sg_publish_data"]
             name_str = sg_data.get("name") or "Unnamed"
             version_number = sg_data.get("version_number")
+            vers_str = "%03d" % version_number if version_number else "N/A"
```
jfboismenu: same comment about version 0.

python/tk_multi_loader/model_latestpublish.py
```diff
@@ -211,8 +211,11 @@ def _set_tooltip(self, item, sg_item):
         else:
             author_str = "Unspecified User"
 
-        tooltip += "<br><br><b>Version:</b> %03d by %s at %s" % (
-            sg_item.get("version_number"),
+        version = sg_item.get("version_number")
+        vers_str = "%03d" % version if version else "N/A"
+
+        tooltip += "<br><br><b>Version:</b> %s by %s at %s" % (
```
jfboismenu: Again, v0

hooks/tk-flame_actions.py
krbg: I catch FlameActionError because theses errors are not "bad" errors and I don't want to print the stack trace in the console. Every other errors that might happens are catched by the loader app.

jfboismenu: Could we log a warning instead then? Having 10 pop-ups would be really annoying, to be honest.

krbg:
@jfboismenu I get rid of the queries on `sg_.*`

hooks/tk-flame_actions.py
jfboismenu: After discussing on Slack, it appears I was completely wrong. The entity being passed here are Shots and not publishes. This is because this is an entity (Shot, Asset, Project, etc...) hook and not a action (PublishedFile) hook.

Please add some comments so that the next person who stumbles on this code knows that those are Shots and not PublishedFiles.

PR #52, opened 2017-07-24
SMOK-29361 | tk-flame_action improved

krbg:
DESCRIPTION

This adds functionality to the tk-flame part of the Shotgun Loader

Now a user can Create a batch group using the supported clips linked to
a Shot entity. Once this Batch group is created, if the user have the
right context (Classic integration for now with the tk-flame-export),
a Render will trigger a publish export to Shotgun.

There's also more configuration options. Theses options allow the user
to change certain behavior of the loader without having to play in the
Python code.

Finally, there's more documentation and more debug logging to be able to
do better troubleshooting.

### DIFF
--- a/hooks/tk-flame_actions.py
+++ b/hooks/tk-flame_actions.py
@@ -13,7 +13,6 @@
 Hook that loads defines all the available actions, broken down by publish type.
 """
 import collections
-import ntpath
 import os
 import re
 
@@ -24,13 +23,7 @@ from sgtk import TankError
 if sgtk.platform.current_engine().is_version_less_than("2018.2"):
     flame = None
 else:
-    try:
-        # The python API is available
-        import flame
-
-    except ImportError:
-        # Unable to load the python API
-        flame = None
+    import flame
 
 HookBaseClass = sgtk.get_hook_baseclass()
 
@@ -43,7 +36,7 @@ SHOT_LOAD_ACTION = "load_batch"
 SHOT_CREATE_ACTION = "create_batch"
 
 
-class FlameActionsError(Exception):
+class FlameActionError(Exception):
     pass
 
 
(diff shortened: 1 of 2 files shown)


hooks/tk-flame_actions.py
```diff
@@ -142,103 +167,127 @@ def execute_action(self, name, params, sg_publish_data):
...
 
-        # Only load the batch if it exists
-        if batch_path and os.path.exists(batch_path):
+        # Directly load the setup if exists
+        if setup_path and os.path.exists(setup_path):
             flame.batch.go_to()
-            flame.batch.load_setup(batch_path)
-            flame.batch.organize()
+
+            if not flame.batch.load_setup(setup_path):
```
jpbrault: Does the flame python API return error?

I'm pretty sure the flame code will check for setup existence when opening the file.
Would be faster to avoid stat'ing the setup file at every layer.

hooks/tk-flame_actions.py
```diff
@@ -247,114 +296,467 @@ def _import_batch_group_from_shot(self, sg_publish_data):
...
+        flame.batch.go_to()
+
+        if not flame.batch.create_batch_group(**batch_group_info):
+            raise FlameActionsError("Unable to create a Batch Group")
+
+        have_write_file = False
+
+        # Import every clips!
+        for clip in clips:
+            if self._exists(clip["path"]):
```
jpbrault: Same comment here, is it worth checking that file exist since it was returned from a list and will be checked again in flame code.

jpbrault: Could check existence of the file only if you fail to import it.

hooks/tk-flame_actions.py
```diff
@@ -378,7 +782,273 @@ def _get_batch_path_from_published_files(self, sg_info):
...
+        """
+
+        # Check if the path exists
+        if os.path.exists(media_path):
+            return True
+
+        folder, file_name = ntpath.split(media_path)
+
+        # Try to check if the path is sa sequence
+        match = re.match(r"(.+\.)((?:\[\d+-\d+\])|(?:\d+))(\..+)", file_name)
```
jpbrault: Would be better to use the MIO::FilePattern code exposed thru boost::python instead.

jpbrault: (Exercise left to the reader :D )

hooks/tk-flame_actions.py
krbg: It’s true. These checks were from the time when the flame api was not returning an error.

krbg: Update: These checks are because if one of the python API command fail, the execution of the python hook that call that command stop. It‘s a weird bug...

manneohrstrom:
@josh-t can you take over this CR since i am about to go on vacation? I'll explain the details around the approach! :)

hooks/tk-flame_actions.py
```diff
@@ -8,24 +8,32 @@
...
 HookBaseClass = sgtk.get_hook_baseclass()
 
 ##############################################################################################################
 # Constants to be used with Flame
 
-# Defines the Schematic Reel we will use
-# Ideally this wouldn't be hardcoded, but there exists no way currently
-# to import clips without specifying a schematic reel
-SCHEMATIC_REEL = "Schematic Reel 1"
+BATCH_ACTION = "load_setup"
```
josh-t: I'm sure I just don't know enough about how flame works, but there are 2 constants with `_batch` in the name, yet `load_setup` is the `BATCH_ACTION`. Seems confusing from an outside perspective.

hooks/tk-flame_actions.py
```diff
@@ -249,125 +281,446 @@ def _import_batch_group_from_shot(self, sg_publish_data):
...
+
+    @property
+    def media_path_root(self):
+        """
+        This path helps to setup the write_file_node by specifying where to write the medias.
+
+        :return: Media path root
+        :type: str
+        """
+        return "/var/tmp"
```
josh-t: This works, but some of these properties could also be done as a class attributes. Would make it easier for subclasses to override.

```python
class FlameActions(HookBaseClass):

    media_path_root = "/var/tmp"
    media_path_pattern = "<shot name>_{segment_name}_v<version>.<frame>"
    media_file_type = "OpenEXR"
    ...
```

Just a thought.

PR #55, opened 2017-10-05
For #44569: Implements file name parsing to determine start and end frame numbers.

thebeeland:
This is resolving an issue in zero config projects. The frame range parsing logic on load in Nuke uses templates to parse the file name. Since we do not have access to templates in a zero config setup, we are forced to instead parse the file name based on some commonly-used conventions.

Note: Nuke does its own frame range parsing when browsing for a frame sequence on disk from the browse button in a read node, however, this functionality is not publicly available via Python. After digging around forums a bit, it looks like people implement their own solutions in Python similar to what I did here. In fact, Foundry provides a tutorial for implementing a managed read node and in that tutorial they provide an example implementation that determines the start/end frames for frame sequences. This seems like the way to go, as a result.

### DIFF
--- a/hooks/tk-nuke_actions.py
+++ b/hooks/tk-nuke_actions.py
@@ -13,6 +13,8 @@ Hook that loads defines all the available actions, broken down by publish type.
 """
 import sgtk
 import os
+import re
+import glob
 
 HookBaseClass = sgtk.get_hook_baseclass()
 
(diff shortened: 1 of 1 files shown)


hooks/tk-nuke_actions.py
```diff
@@ -228,6 +230,57 @@ def _create_read_node(self, path, sg_publish_data):
...
+        #
+        # The number of digits or hashes does not matter; we match as many as
+        # exist.
+        frame_pattern = re.compile(r"([0-9#]+|[%]0\dd)$")
+        root, ext = os.path.splitext(path)
+        match = re.search(frame_pattern, root)
+
+        # If we did not match, we don't know how to parse the file name, or there
+        # is no frame number to extract.
+        if match is None:
```
josh-t: suggest `if not match:`

hooks/tk-nuke_actions.py
```diff
@@ -228,6 +230,57 @@ def _create_read_node(self, path, sg_publish_data):
...
+        # match object:
+        #
+        # 0001
+        # ####
+        # %04d
+        #
+        # The number of digits or hashes does not matter; we match as many as
+        # exist.
+        frame_pattern = re.compile(r"([0-9#]+|[%]0\dd)$")
+        root, ext = os.path.splitext(path)
```
josh-t: Would it be better to split the path from the filename and only operate on the filename? This could help eliminate version number matches in the path itself. Maybe should include `.` separators before and after the frame number token? The `path_info` hook in `tk-multi-publish2` parses frame numbers this way. Not saying it's right for this case, but just for reference: https://github.com/shotgunsoftware/tk-multi-publish2/blob/master/hooks/path_info.py#L30

josh-t: Ah, I see the `$` anchor now. We can continue discussion of centralizing similar logic outside of this PR. No reason to block this. 👍

PR #57, opened 2017-12-15
#45167 Fix for utf-8 file problem in various DCCs

hawif:
Fix problem encountered with utf-8 filenames with various DCCs

### DIFF
--- a/hooks/tk-3dsmaxplus_actions.py
+++ b/hooks/tk-3dsmaxplus_actions.py
@@ -169,14 +169,14 @@ class MaxActions(HookBaseClass):
             )
         )
     
-    def _merge(self, path, sg_publish_data):
+    def _merge(self, path_utf8, sg_publish_data):
         """
         Merge contents of the given file into the scene.
         
         :param path: Path to file.
         :param sg_publish_data: Shotgun data dictionary with all the standard publish fields.
         """
-
+        path = path_utf8.decode('utf-8')
         if not os.path.exists(path):
             raise Exception("File not found on disk - '%s'" % path)
         
--- a/hooks/tk-3dsmaxplus_actions.py
+++ b/hooks/tk-3dsmaxplus_actions.py
@@ -190,7 +190,7 @@ class MaxActions(HookBaseClass):
         app = self.parent
 
         # Note: MaxPlus.FileManager.Merge() is not equivalent as it opens a dialog.
-        app.engine.safe_dialog_exec(lambda: MaxPlus.Core.EvalMAXScript('mergeMAXFile(\"' + path.replace('\\', '/') + '\")'))
+        app.engine.safe_dialog_exec(lambda: MaxPlus.Core.EvalMAXScript('mergeMAXFile(\"' + path_utf8.replace('\\', '/') + '\")'))
 
     def _xref_scene(self, path, sg_publish_data):
         """
@@ -200,10 +200,10 @@ class MaxActions(HookBaseClass):
         :param sg_publish_data: Shotgun data dictionary with all the standard publish fields.
         """
         
-        if not os.path.exists(path):
+        if not os.path.exists(path.decode('utf-8')):
             raise Exception("File not found on disk - '%s'" % path)
         
-        (_, ext) = os.path.splitext(path)
+        (_, ext) = os.path.splitext(path.decode('utf-8'))
         
         supported_file_exts = [".max"]
         if ext.lower() not in supported_file_exts:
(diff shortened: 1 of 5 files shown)


hooks/tk-3dsmaxplus_actions.py
```diff
@@ -169,14 +169,14 @@ def _import_alembic(self, path):
             )
         )
     
-    def _merge(self, path, sg_publish_data):
+    def _merge(self, path_utf8, sg_publish_data):
```
manneohrstrom: nice and clear!

hooks/tk-3dsmaxplus_actions.py
```diff
@@ -169,14 +169,14 @@ def _import_alembic(self, path):
...
         """
         Merge contents of the given file into the scene.
         
         :param path: Path to file.
```
manneohrstrom: please update doc strings to reflect your variable name change.

hooks/tk-3dsmaxplus_actions.py
```diff
@@ -169,14 +169,14 @@ def _import_alembic(self, path):
...
         :param sg_publish_data: Shotgun data dictionary with all the standard publish fields.
         """
-
+        path = path_utf8.decode('utf-8')
```
manneohrstrom: please use double quotes for strings in the tk code base.

manneohrstrom: since unicode is so tricky, let's make sure we add plenty of comments to explain what is going on. I'll just comment for 3dsmax in the pull request but please have a look at what's appropriate for each DCC below.

Please add a comment explaining why we are doing this and what's going on. I suggest something like this:

```
# toolkit uses utf-8 encoded strings internally and the 3dsmax API expects unicode
# so convert the path to ensure filenames containing complex characters are supported
```

hooks/tk-3dsmaxplus_actions.py
manneohrstrom: questions:

- why are we not doing this for alembic?
- to avoid repetition could we do the conversion in `execute_action` above (after line 135)?
- if this is not the case, we need some clear comments explaining this, to help future readers and maintainers of the code. Right now it's not quite clear why the code is laid out the way it is.

hooks/tk-3dsmaxplus_actions.py
```diff
@@ -200,10 +200,10 @@ def _xref_scene(self, path, sg_publish_data):
         :param sg_publish_data: Shotgun data dictionary with all the standard publish fields.
         """
         
-        if not os.path.exists(path):
+        if not os.path.exists(path.decode('utf-8')):
```
manneohrstrom: lots of `.decode()` repetition here. why not convert path to unicode only once?

hooks/tk-maya_actions.py
```diff
@@ -243,7 +243,7 @@ def _create_udim_texture_node(self, path, sg_publish_data):
             mel.eval("generateUvTilePreview %s" % file_node)
         return file_node
 
-    def _create_image_plane(self, path, sg_publish_data):
+    def _create_image_plane(self, path_utf8, sg_publish_data):
```
manneohrstrom: same here - if image planes require different unicode input than udim texture nodes and other maya things, we should add some really clear comments in the code explaing this. If not, we should lift up the conversion a level (line 157), so we only do it once for all the maya logic

hooks/tk-nuke_actions.py
```diff
@@ -147,14 +147,15 @@ def execute_action(self, name, params, sg_publish_data):
...
-    def _import_script(self, path, sg_publish_data):
+    def _import_script(self, path_utf8, sg_publish_data):
         """
         Import contents of the given file into the scene.
         
         :param path: Path to file.
         :param sg_publish_data: Shotgun data dictionary with all the standard publish fields.
         """
         import nuke
+        path=path_utf8.decode('utf-8')
```
manneohrstrom: can this be nuke generic? same question as for maya and max above?

hooks/tk-photoshopcc_actions.py
```diff
@@ -161,14 +161,15 @@ def _open_file(self, path, sg_publish_data):
...
-    def _place_file(self, path, sg_publish_data):
+    def _place_file(self, path_utf8, sg_publish_data):
         """
         Import contents of the given file into the scene.
 
         :param path: Path to file.
         :param sg_publish_data: Shotgun data dictionary with all the standard
                                 publish fields.
         """
+        path=path_utf8.decode('utf-8')
```
manneohrstrom: can this be ps generic? same question as for maya and max above?

PR #58, opened 2018-01-11
#37225 Fixed renamed entity display name

hawif:
Fixed renamed entity display name in loader app


manneohrstrom:
this solution is a workaround for the actual issue - which is the fact that the sg globals never invalidates the schema caches. I am going to close this and we should fix this in shotgunglobals instead.

PR #59, opened 2018-01-16
#45165 fixed utf8 file encoding problem

hawif:
Fixed utf8 encoding problem when creating read node and importing script in nuke.

### DIFF
--- a/hooks/tk-nuke_actions.py
+++ b/hooks/tk-nuke_actions.py
@@ -133,8 +133,6 @@ class NukeActions(HookBaseClass):
                       "Parameters: %s. Publish Data: %s" % (name, params, sg_publish_data))
         
         # resolve path - forward slashes on all platforms in Nuke
-        # toolkit uses utf-8 encoded strings internally and Nuke API expects unicode
-        # so convert the path to ensure filenames containing complex characters are supported
         path = self.get_publish_path(sg_publish_data).replace(os.path.sep, "/")
         
         if name == "read_node":
@@ -157,6 +155,8 @@ class NukeActions(HookBaseClass):
         :param sg_publish_data: Shotgun data dictionary with all the standard publish fields.
         """
         import nuke
+
+        # must use unicode otherwise path won't be found
         if not os.path.exists(path.decode('utf-8')):
             raise Exception("File not found on disk - '%s'" % path)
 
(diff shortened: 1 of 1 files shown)


hooks/tk-nuke_actions.py
```diff
@@ -160,7 +160,7 @@ def _import_script(self, path, sg_publish_data):
         if not os.path.exists(path):
             raise Exception("File not found on disk - '%s'" % path)
 
-        nuke.nodePaste(path)
+        nuke.nodePaste(path.encode('utf-8'))
```
manneohrstrom: I am missing a comment here and it seems this is the wrong way around again.
Am I right thinking that we are turning a unicode path into `utf-8` encoded string of bytes?
Why is path a unicode string? I think the issue we should fix is upstream - like i have pointed out in the past, we want things to be utf-8 for as long as possible in toolkit and if they need to be converted to unicode objects, we do that at the "bondary" of toolkit, right before it leaves the system and enters another. 

On line 138 i see this line:

```
path = self.get_publish_path(sg_publish_data).replace(os.path.sep, "/").decode("utf-8")
```

So we are decoding it, only to encode it a few lines later down in the code? This doesn't look right to me.

hawif: In all other DCCs, we are decoding path at this level. I noticed that at some places that come afterwards we still need utf-8 encoded strings, but we can go the other way.

manneohrstrom: why can't you just remove the `.decode("utf-8")` suffix on line 138 and then you don't need any of the encoding in `_create_read_node`, `_import_script` and `_open_project`?

hawif: yes, I was testing that, it works, but we still need to place a decode in import_script

hooks/tk-nuke_actions.py
```diff
@@ -157,7 +157,7 @@ def _import_script(self, path, sg_publish_data):
         :param sg_publish_data: Shotgun data dictionary with all the standard publish fields.
         """
         import nuke
-        if not os.path.exists(path):
+        if not os.path.exists(path.decode('utf-8')):
```
manneohrstrom: please add a comment pointing out that this is required for unicode paths on windows, however the nuke API itself handles utf-8 encoded paths correctly.

PR #60, opened 2018-01-17
Create reference using relative path instead of absolute path in Maya

qpskcn1:
When we open up an empty Maya scene and load some assets published through different OS, everything looks fine. 
But if we open a Maya scene which have references created by loader under other OS, the path would be wrong.
For example, there is an asset located at our network storage, we use loader to create a reference in a Maya scene. At this moment, the reference we created uses path `/mnt/assets/brutus`. Then we open this Maya scene in Windows, this reference would be missed since the path of that reference cannot be found.

Using relative path could resolve this issue.

### DIFF
--- a/hooks/tk-maya_actions.py
+++ b/hooks/tk-maya_actions.py
@@ -192,8 +192,11 @@ class MayaActions(HookBaseClass):
         namespace = "%s %s" % (sg_publish_data.get("entity").get("name"), sg_publish_data.get("name"))
         namespace = namespace.replace(" ", "_")
                 
-        pm.system.createReference(path, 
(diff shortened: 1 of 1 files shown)


manneohrstrom:
Hi! Thanks for the PR! Heads up that toolkit might already do what you are looking for, without having to make the paths local. Check out https://support.shotgunsoftware.com/hc/en-us/articles/115000067493-Integrations-Admin-Guide#Resolving%20file%20URLs

If you set the env vars `SHOTGUN_PATH_WINDOWS`, `SHOTGUN_PATH_MAC`, and `SHOTGUN_PATH_LINUX`, Toolkit should be able to translate paths cross platform automatically!

PR #62, opened 2018-01-18
#45167 removed unicode flag from regex sub operation

hawif:
the Unicode flag was causing problems in regex sub operation. Houdini does not accept special characters (é, à,..) when creating nodes. We replace these by '_', along with non alphanumeric characters.

### DIFF
--- a/hooks/tk-houdini_actions.py
+++ b/hooks/tk-houdini_actions.py
@@ -231,7 +231,9 @@ class HoudiniActions(HookBaseClass):
         # remove non alphanumeric characers from the string (houdini node names
         # must be alphanumeric). first, build a regex to match non alpha-numeric
         # characters. Then use it to replace any matches with an underscore
-        pattern = re.compile('[\W_]+', re.UNICODE)
+        
+        # cannot use special characters to create nodes
+        pattern = re.compile('[\W_]+')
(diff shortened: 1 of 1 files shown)


hawif:
image
This commit solves the problem of creating a new node when executing COP command. But there is still a problem in the info section of the node.

hooks/tk-houdini_actions.py
```diff
@@ -231,7 +231,9 @@ def _file_cop(self, path, sg_publish_data):
         # remove non alphanumeric characers from the string (houdini node names
         # must be alphanumeric). first, build a regex to match non alpha-numeric
         # characters. Then use it to replace any matches with an underscore
-        pattern = re.compile('[\W_]+', re.UNICODE)
+        
+        # cannot use special characters to create nodes
+        pattern = re.compile('[\W_]+')
```
manneohrstrom: this change looks right to me - `path` should always be a utf-8 encoded string, so the `re.UNICODE` directive that got removed didn't make any sense!

josh-t: looks good. Unrelated, I wish this code would do something like `achmédèád.jpg` => `achmedead.jpg` instead of `achm_d_.jpg`

PR #64, opened 2018-02-16
Multiselection and hierarchical Tag entity support

reformstudios:
The purpose of this pull request is to allow studios to use the Tag system to browse a media/asset library.

Two major updates to tk-multi-loader2 were required to achieve this.
- Allow multiselection in the tree-view. (along with various updates to functions to support list items rather than single items)
- Add support for the 'name' field of 'Tag' entities.

No additional entities are required, but a new field for Tag entities should be added to allow a hierarchy to be formed. The field should be called 'Parent Tag' (or 'sg_parent_tag' in code). It allows you to group tags under other tags to create hierarchies and to organise tags by type or purpose.
For the purpose of a studio library, a new Tag should be created to be the root Tag for the library tag hierarchy; eg 'Library Tag'.

To add a Tag tab to your loader UI, use the following entry in your loader2 config.
```
entities:
        - caption: Asset Library
          hierarchy: [sg_parent_tag, name]
          entity_type: Tag
          publish_filters:
          - [project, is, {type: Project, id: 90}] # THIS ID SHOULD POINT TO THE PROJECT HOLDING YOUR STUDIO LIBRARY MEDIA/ASSETS.
          filters:
          - [sg_parent_tag.Tag.sg_parent_tag, is, {type: Tag, id: 374}] # THIS ID SHOULD POINT TO THE 'LIBRARY TAG' THAT YOU CREATED PREVIOUSLY.

```

You can then create a hierarchy of new Tags that define the top groupings of tag for your library. These should all have their parent sent to the 'library tag' tag. The hierarchy can only be 1 level deep currently
eg.
- Type
- - 2D Element
- - 3D Asset
- - Texture
- Subject
- - Blood
- - Dust
- - Bubbles
- - Debris
- - MuzzleFlash
- Style
- - explosion
- - backfire
- - spurt
- - splatter
- - demolition
- Vehicle
- - Car
- - Bus
- - Spaceship
- Location
- - London
- - Paris
- - New York
- - Desert
- - Space
... etc

The multiselection behaviour in the tree is Toggle behaviour, with each additional item selected narrowing down the number of published items displayed in the right-hand panel. This allows users to use an arbitrary number and type of tags to hone in on relevant search results.

### DIFF
--- a/python/tk_multi_loader/dialog.py
+++ b/python/tk_multi_loader/dialog.py
@@ -1680,7 +1680,7 @@ class AppDialog(QtGui.QWidget):
             self._setup_details_panel([])
 
             # tell the publish view to change
-            self._load_publishes_for_entity_item(selected_item)
+            self._load_publishes_for_entity_items([selected_item])
 
     def _on_treeview_item_selected(self):
         """
@@ -1716,82 +1716,6 @@ class AppDialog(QtGui.QWidget):
 
         # tell publish UI to update itself
         self._load_publishes_for_entity_items(selected_items)
-
-    def _load_publishes_for_entity_item(self, item):
-        """
-        Given an item from the treeview, or None if no item
-        is selected, prepare the publish area UI.
-        """
-        # clear selection. If we don't clear the model at this point,
-        # the selection model will attempt to pair up with the model is
-        # data is being loaded in, resulting in many many events
-        self.ui.publish_view.selectionModel().clear()
-
-        # Determine the child folders.
-        child_folders = []
-        proxy_model = self._entity_presets[self._current_entity_preset].proxy_model
-
-        if item is None:
-            # nothing is selected, bring in all the top level
-            # objects in the current tab
-            num_children = proxy_model.rowCount()
-
-            for x in range(num_children):
-                # get the (proxy model) index for the child
-                child_idx_proxy = proxy_model.index(x, 0)
-                # switch to shotgun model index
-                child_idx = proxy_model.mapToSource(child_idx_proxy)
-                # resolve the index into an actual standarditem object
-                i = self._entity_presets[self._current_entity_preset].model.itemFromIndex(child_idx)
-                child_folders.append(i)
-
-        else:
-            # we got a specific item to process!
-
-            # now get the proxy model level item instead - this way we can take search into
-            # account as we show the folder listings.
-            root_model_idx = item.index()
-            root_model_idx_proxy = proxy_model.mapFromSource(root_model_idx)
-            num_children = proxy_model.rowCount(root_model_idx_proxy)
-
-            # get all the folder children - these need to be displayed
-            # by the model as folders
-
-            for x in range(num_children):
-                # get the (proxy model) index for the child
-                child_idx_proxy = root_model_idx_proxy.child(x, 0)
-                # switch to shotgun model index
-                child_idx = proxy_model.mapToSource(child_idx_proxy)
-                # resolve the index into an actual standarditem object
-                i = self._entity_presets[self._current_entity_preset].model.itemFromIndex(child_idx)
-                child_folders.append(i)
-
-        # Is the show child folders checked?
-        # The hierarchy model cannot handle "Show items in subfolders" mode.
-        show_sub_items = self.ui.show_sub_items.isChecked() and \
-                         not isinstance(self._entity_presets[self._current_entity_preset].model, SgHierarchyModel)
-
-        if show_sub_items:
-            # indicate this with a special background color
-            self.ui.publish_view.setStyleSheet("#publish_view { background-color: rgba(44, 147, 226, 20%); }")
-            if len(child_folders) > 0:
-                # delegates are rendered in a special way
-                # if we are on a non-leaf node in the tree (e.g there are subfolders)
-                self._publish_thumb_delegate.set_sub_items_mode(True)
-                self._publish_list_delegate.set_sub_items_mode(True)
-            else:
-                # we are at leaf level and the subitems check box is checked
-                # render the cells
-                self._publish_thumb_delegate.set_sub_items_mode(False)
-                self._publish_list_delegate.set_sub_items_mode(False)
-        else:
-            self.ui.publish_view.setStyleSheet("")
-            self._publish_thumb_delegate.set_sub_items_mode(False)
-            self._publish_list_delegate.set_sub_items_mode(False)
-
-        # now finally load up the data in the publish model
-        publish_filters = self._entity_presets[self._current_entity_preset].publish_filters
-        self._publish_model.load_data(item, child_folders, show_sub_items, publish_filters)
 
     def _load_publishes_for_entity_items(self, items):
         """
--- a/python/tk_multi_loader/dialog.py
+++ b/python/tk_multi_loader/dialog.py
@@ -1233,9 +1233,7 @@ class AppDialog(QtGui.QWidget):
                 sg_entity_type = setting_dict["entity_type"]
 
             # Check to see if we are showing a tags view.
-            # Library tags are a new entity so have a temporary entity name of "CustomEntity14"
-            # I suggest we create a new project specific tag field as part of a default SG project called 'project_tag'
-            if sg_entity_type == 'CustomEntity14':
+            if sg_entity_type == 'Tag':
                 type_tag = True
             else:
                 type_tag = False
@@ -1276,10 +1274,8 @@ class AppDialog(QtGui.QWidget):
 
             # Enable multiselection on tag list entities (or in this case extended selection so selections are sticky)
             if type_tag:
-                self._log_info('view.setSelectionMode(QtGui.QAbstractItemView.ExtendedSelection)')
                 view.setSelectionMode(QtGui.QAbstractItemView.MultiSelection)
             else:
-                self._log_info('view.setSelectionMode(QtGui.QAbstractItemView.SingleSelection)')
                 view.setSelectionMode(QtGui.QAbstractItemView.SingleSelection)
 
             # Keep a handle to all the new Qt objects, otherwise the GC may not work.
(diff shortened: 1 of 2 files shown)


python/tk_multi_loader/model_latestpublish.py
```diff
@@ -66,7 +66,7 @@ def get_associated_tree_view_item(self, item):
         entity_item_hash = item.data(self.ASSOCIATED_TREE_VIEW_ITEM_ROLE)
         return self._associated_items.get(entity_item_hash)
 
-    def load_data(self, item, child_folders, show_sub_items, additional_sg_filters):
+    def load_data(self, items, child_folders, show_sub_items, additional_sg_filters):
```
josh-t: i don't think changing this param will cause any issues assuming all calling code has been updated, but should definitely update the docstring to reflect.

python/tk_multi_loader/model_latestpublish.py
```diff
@@ -79,107 +79,129 @@ def load_data(self, item, child_folders, show_sub_items, additional_sg_filters):
...
-        if item is None:
+        if len(items) == 0:
             # nothing selected in the treeview
             # passing none to _load_data indicates that no query should be executed
-            sg_filters = None
+            sg_filters = []
 
-        else:
-            # we have a selection!
+        else: #if isinstance(items, (list, tuple)):
```
josh-t: suggest removing this commented out condition

python/tk_multi_loader/model_latestpublish.py
```diff
@@ -79,107 +79,129 @@ def load_data(self, item, child_folders, show_sub_items, additional_sg_filters):
...
+                    # now create the final query for the model - this will be
+                    # a big in statement listing all the ids returned from
+                    # the previous query, asking the model to only show the
+                    # items matching the previous query.
+                    #
+                    # note that for tasks, we link via the task field
+                    # rather than the std entity link field
+                    #
+
+                    # New sg_filter for tags. We need to pull the tag applied
```
josh-t: suggest removing the "New " here and treat this as a note like the comment above.

python/tk_multi_loader/model_latestpublish.py
```diff
@@ -79,107 +79,129 @@ def load_data(self, item, child_folders, show_sub_items, additional_sg_filters):
...
+                    # to the Version associated with the publish
+                    # In the context of a media library it should be assumed
+                    # that any PublishedFile WILL have a Version associated with it.
+                    # We may need to add logic to cover cases where the published file has no version.
+                    if entity_type == "Task":
+                        sg_filters = [["task", "in", data]]
+                    elif entity_type == "Version":
+                        sg_filters = [["version", "in", data]]
+                    elif entity_type == "Tag":
+                        sg_filters = [["version.Version.tags", "in", data ]]
```
josh-t: I'm wondering if we want to expose this field in the config settings instead. for example, could i configure the tag matching for tags on the publish's linked entity instead? what do you think?

python/tk_multi_loader/model_latestpublish.py
```diff
@@ -79,107 +79,129 @@ def load_data(self, item, child_folders, show_sub_items, additional_sg_filters):
...
+                            # handle tags via the publish file's Version's tags field. (Tags should always be applied
+                            # to the Version rather than the PublishedFile as we also want to be able to filter by tag
+                            # from the media page (which is shows only versions and as the published_files field in
+                            # versions is ulti-entity, we cant filter by published_files.PublishedFile.tag)
+                            if sg_data.get("type") == "Task":
+                                sg_filters.append(["task", "is", {"type": sg_data["type"], "id": sg_data["id"]} ])
+                            elif sg_data.get("type") == "Version":
+                                sg_filters = [["version", "is", {"type": "Version", "id": sg_data["id"]}]]
+                            elif sg_data.get("type") == "Tag":
+                                sg_filters.append(["version.Version.tags", "in", {"type": sg_data["type"], "id": sg_data["id"]} ])
```
josh-t: same comment here about exposing this field in the app's settings.

python/tk_multi_loader/model_latestpublish.py
```diff
@@ -79,107 +79,129 @@ def load_data(self, item, child_folders, show_sub_items, additional_sg_filters):
...
+                        # Extract the Shotgun data and field value from the node item.
+                        (sg_data, field_value) = model_item_data.get_item_data(item)
+
+                        if sg_data:
+                            # leaf node!
+                            # show the items associated. Handle tasks
+                            # via the task field instead of the entity field
+                            # handle tags via the publish file's Version's tags field. (Tags should always be applied
+                            # to the Version rather than the PublishedFile as we also want to be able to filter by tag
+                            # from the media page (which is shows only versions and as the published_files field in
```
josh-t: typos "which is shows" and "ulti-entity"

python/tk_multi_loader/model_latestpublish.py
```diff
@@ -79,107 +79,129 @@ def load_data(self, item, child_folders, show_sub_items, additional_sg_filters):
...
                         sg_filters = None
-
+                    else:
+                        # Extract the Shotgun data and field value from the node item.
+                        (sg_data, field_value) = model_item_data.get_item_data(item)
+
+                        if sg_data:
+                            # leaf node!
+                            # show the items associated. Handle tasks
+                            # via the task field instead of the entity field
```
josh-t: can you add a period to the end of the old comment here and format the lines a bit for consistency.

python/tk_multi_loader/model_latestpublish.py
```diff
@@ -79,107 +79,129 @@ def load_data(self, item, child_folders, show_sub_items, additional_sg_filters):
...
-                        # can have publishes of its own associated
-                        sg_filters = [["entity", "is", field_value ]]
-
-                    else:
-                        # this is an intermediate node like status or asset type which does not
-                        # have any publishes of its own, because the value (e.g. the status or the asset type)
-                        # is nothing that you could link up a publish to.
+                # loop through the selected items
+                for item in items:
+                    if item is None:
```
josh-t: checking for a `None` item here but not in the initial check and `show_sub_item` loops above. is this necessary? if so, do need to account for it above?

python/tk_multi_loader/model_latestpublish.py
```diff
@@ -79,107 +79,129 @@ def load_data(self, item, child_folders, show_sub_items, additional_sg_filters):
...
+                        # nothing selected in the treeview
+                        # passing none to _load_data indicates that no query should be executed
```
josh-t: in the initial check on line `85` above, this was changed to be set to `[]`. Was that necessary? Should this change if so?

python/tk_multi_loader/model_latestpublish.py
```diff
@@ -79,107 +79,129 @@ def load_data(self, item, child_folders, show_sub_items, additional_sg_filters):
...
+                            if isinstance(field_value, dict) and "name" in field_value and "type" in field_value:
+                                # this is an intermediate node like a sequence or an asset which
+                                # can have publishes of its own associated
+                                sg_filters = [["entity", "is", field_value ]]
+
+                            else:
+                                # this is an intermediate node like status or asset type which does not
+                                # have any publishes of its own, because the value (e.g. the status or the asset type)
+                                # is nothing that you could link up a publish to.
+                                sg_filters = None
```
josh-t: same comment here about consistency when this is set to `None` or `[]`

python/tk_multi_loader/model_latestpublish.py
```diff
@@ -536,3 +556,20 @@ def _before_data_processing(self, sg_data_list):
         self._publish_type_model.set_active_types( type_id_aggregates )
 
         return new_sg_data
+
+    def _log_debug(self, msg):
```
josh-t: same comment here about whether these are necessary.

python/tk_multi_loader/dialog.py
```diff
@@ -1242,6 +1249,12 @@ def _load_entity_presets(self):
             view.setHeaderHidden(True)
             view.setModel(proxy_model)
 
+            # Enable multiselection on tag list entities (or in this case extended selection so selections are sticky)
```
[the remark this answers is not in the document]
reformstudios: Sure, multiselect requires shift or ctrl to be pressed. Extended selection doesn't; the selection remains sticky and you can click to toggle items on and off which I thought was a more appropriate UI behaviour for tag filters.

python/tk_multi_loader/dialog.py
```diff
@@ -1643,40 +1657,52 @@ def _switch_profile_tab(self, new_index, track_in_history):
...
 
-        selected_item = self._get_selected_entity()
+        model = self._entity_presets[self._current_entity_preset].model
 
-        # update breadcrumbs
-        self._populate_entity_breadcrumbs(selected_item)
+        # If nothing is selected, refresh the view.
+        if len(selected_items) == 0:
+            if isinstance(model, SgEntityModel):
+                model.async_refresh()
```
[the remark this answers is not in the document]
reformstudios: I could be wrong it's been a while since I looked at this, but if you don't refresh the view when nothing is selected, then you'll be left with a filtered list of published_files, where you actually want the list to reset to fully unfiltered.

python/tk_multi_loader/model_latestpublish.py
reformstudios: True, I don't see why not. 
My use-case is specific; a media library that might be browsed via the loader2 or the website Media pages. As there is no way to filter the website 'Media' views by PublishedFile tag (eg, when you are using the Media view to browse a media library), the the tag must be on the Version.

This is also a structural limitation; there can be multiple published_files with different tags per Version, whereas there can only be one Version per multiple published files. If the tags are on the published files, there is no way to filter a list of Versions by any publishedFile fields is there?

It therefore, in my view, makes no sense to tag published files as the functionality would not extend beyond the workfiles app. At least, certainly from the use-case of a media library. Which is the key purpose of introducing support for Tag entities. 

Other use-cases might indeed benefit, eg where you don't have a Version associated with a published file, or where you need to tag publishes of a single Version with different tags....

Anyway, I'm all for leaving things as configurable as possible, so we should just do what you suggest. Can you tell me how to do that?

dalorin:
Just came across this pull request while trying to implement the exact same functionality. Any chance of this coming back to life and being merged to master?

PR #65, opened 2018-03-07
Added Mp4 to Valid Extensions

Masepack:
Allows published .mp4's to be loaded as a Read Node in Nuke.


manneohrstrom:
Looks good @Masepack let me know if you need info about our process conventions around merging and QA etc!

PR #66, opened 2018-04-10
Load image file as Clip in NukeStudio

Acidpix:


thebeeland:
Hey there! Thanks a ton for sending the PR our way. I'm digging into this now, as well as your PR for the same workflow in tk-multi-breakdown. We're interested in taking both with some minor cleanup that I'll do before getting the changes onto master. Once I've done that, I'll close the PRs out.

thebeeland:
We have this in QA now as part of #68. I'll close this PR now in favor of that one! Thanks a ton for sending this over to us, it's a great idea.

PR #67, opened 2018-04-12
SG-5399 Added support for complex filters

pscadding:
Now allows complex filters to be used when specifying the filters for tabs.
For example the filters in the tab settings bellow would not work previous to the update.

```
- caption: Tasks
    entity_type: Task
    filters:
      - [project, is, "{context.project}"]
      - filter_operator: "any"
        filters:
          - [entity, type_is, Asset]
          - [entity, type_is, Shot]
```

I have taken the code from the [workfiles2 way of resolving the filters](https://github.com/shotgunsoftware/tk-multi-workfiles2/blob/master/python/tk_multi_workfiles/util.py#L282-L314). I've also moved the logic for resolving filters out into utils, as this is what wf2 does as well.

However I also noticed that the workfiles way of doing it doesn't resolve the project from the context, so I've made a hybrid of the two bits of logic.

For SG-5399

### DIFF
--- a/python/tk_multi_loader/utils.py
+++ b/python/tk_multi_loader/utils.py
@@ -281,12 +281,12 @@ def filter_publishes(app, sg_data_list):
 
 def resolve_filters(filters):
     """
+    When passed a list of filters, it will resolve strings found in the filters using the context.
+    For example: '{context.user}' could get resolved to {'type': 'HumanUser', 'id': 86, 'name': 'Philip Scadding'}
 
-    When passed a list of filters, it will resolve strings found in the filters using the context
-    example: '{context.user}' could get resolved to {'type': 'HumanUser', 'id': 86, 'name': 'Philip Scadding'}
-
-    :param filters: a list of filters as found in the info.yml config
-    should be in the format: [[task_assignees, is, '{context.user}'],[sg_status_list, not_in, [fin,omt]]]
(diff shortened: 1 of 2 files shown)


python/tk_multi_loader/utils.py
```diff
@@ -276,3 +277,45 @@ def filter_publishes(app, sg_data_list):
         sg_data_list = []
 
     return sg_data_list
+
+
+def resolve_filters(filters):
+    """
+
```
manneohrstrom: Please remove the whitespace line here.

python/tk_multi_loader/utils.py
```diff
@@ -276,3 +277,45 @@ def filter_publishes(app, sg_data_list):
...
+    When passed a list of filters, it will resolve strings found in the filters using the context
```
manneohrstrom: nitpick: please add period at the end

python/tk_multi_loader/utils.py
```diff
@@ -276,3 +277,45 @@ def filter_publishes(app, sg_data_list):
...
+    example: '{context.user}' could get resolved to {'type': 'HumanUser', 'id': 86, 'name': 'Philip Scadding'}
```
manneohrstrom: example -> For example:

PR #73, opened 2019-06-03
Removed use of PyMEL

UnaiM:


UnaiM:
Closing as it’s not needed— the hook can be overrided by a PyMEL-less variant in the config already.

PR #74, opened 2019-06-05
Making all signal connections use partials rather than lambdas to wor…

Liametc:
…k with katana

In katana, signals that are connected to lambdas return booleans rather than the desired objects. Using partials fixes this behaviour

### DIFF
--- a/python/tk_multi_loader/loader_action_manager.py
+++ b/python/tk_multi_loader/loader_action_manager.py
@@ -10,6 +10,7 @@
 
 import sgtk
 import datetime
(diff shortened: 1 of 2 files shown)


Liametc:
@thebeeland tagging you (again, sorry) to also get this logged. Goes with [this](https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/93) PR as part of getting shotgun into Katana

pscadding:
No problem I've logged this one as well, thanks!

PR #75, opened 2019-07-27
Add publish_extra_details hook

000paradox000:
In the details pane, puiblish history version and publish list of this app would be great to have the chance for adding extra fields like status of the published file.

A nice way to archieve this, is using a HOOK. We propose to add the hook publish_extra_details which receives a sg_item and optionally a kwargs dict and returns a tuple of tuples with the format (Label, Value).

This hook applies only for published files.

Visual example:

image

### DIFF
--- a/hooks/publish_extra_details.py
+++ b/hooks/publish_extra_details.py
@@ -8,9 +8,6 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
-import os
-
-import sgtk
 from sgtk import Hook
 
 
--- a/hooks/publish_extra_details.py
+++ b/hooks/publish_extra_details.py
@@ -8,17 +8,20 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
-from sgtk import Hook
+import sgtk
+
+HookBaseClass = sgtk.get_hook_baseclass()
 
 
-class PublishExtraDetails(Hook):
+class PublishExtraDetails(HookBaseClass):
     """Hook that can be used to add extra fields in the details pane."""
 
-    def execute(self, sg_item, **kwargs):
+    def execute(self, entity):
         """
         Main hook entry point.
 
-        :param sg_item: published item
-        :return List: Tuple of tuples with the form (Label, Value)
+        :param entity: The entity dictionary associated with the published item to be displayed in the loader app.
+        :return: A dictionary in the form of {label: value}
+        :rtype: dictionary
         """
-        return ()
+        return {}
--- a/hooks/publish_extra_details.py
+++ b/hooks/publish_extra_details.py
@@ -1,0 +1,27 @@
+# Copyright (c) 2015 Shotgun Software Inc.
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import os
+
+import sgtk
+from sgtk import Hook
+
+
+class PublishExtraDetails(Hook):
+    """Hook that can be used to add extra fields in the details pane."""
+
+    def execute(self, sg_item, **kwargs):
+        """
+        Main hook entry point.
+
+        :param sg_item: published item
+        :return List: Tuple of tuples with the form (Label, Value)
+        """
+        return ()
--- a/info.yml
+++ b/info.yml
@@ -40,7 +40,7 @@ configuration:
     publish_extra_details_hook:
         type: hook
         default_value: "{self}/publish_extra_details.py"
-        description: Hook which contains extra fields to be displayed on the details pane.
+        description: Hook that can be used to return extra fields to be displayed for a publish in the details pane.
 
     download_thumbnails:
         type: bool
--- a/info.yml
+++ b/info.yml
@@ -37,6 +37,11 @@ configuration:
         description: Specify a hook that, if needed, can filter the raw list of publishes returned
                      from Shotgun for the current location.
 
+    publish_extra_details_hook:
+        type: hook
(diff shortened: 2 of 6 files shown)


hooks/publish_extra_details.py
```diff
@@ -0,0 +1,27 @@
...
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import os
+
+import sgtk
```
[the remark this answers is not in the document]
000paradox000: Done

hooks/publish_extra_details.py
```diff
@@ -0,0 +1,27 @@
...
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import os
```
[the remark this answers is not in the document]
000paradox000: Done

hooks/publish_extra_details.py
```diff
@@ -0,0 +1,24 @@
...
+
+class PublishExtraDetails(Hook):
+    """Hook that can be used to add extra fields in the details pane."""
+
+    def execute(self, sg_item, **kwargs):
+        """
+        Main hook entry point.
+
+        :param sg_item: published item
+        :return List: Tuple of tuples with the form (Label, Value)
```
thebeeland: Do this instead:

```
:return: Tuple of tuples in the form of ((label, value), ...)
:rtype: tuple
```

hooks/publish_extra_details.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class PublishExtraDetails(Hook):
+    """Hook that can be used to add extra fields in the details pane."""
+
+    def execute(self, sg_item, **kwargs):
+        """
+        Main hook entry point.
+
+        :param sg_item: published item
```
thebeeland: I'd call this something other than "sg_item". Maybe something like the following?

```
:param entity: The entity dictionary associated with the published item to be displayed in the loader app.
```

info.yml
```diff
@@ -37,6 +37,11 @@ configuration:
         description: Specify a hook that, if needed, can filter the raw list of publishes returned
                      from Shotgun for the current location.
 
+    publish_extra_details_hook:
+        type: hook
+        default_value: "{self}/publish_extra_details.py"
+        description: Hook which contains extra fields to be displayed on the details pane.
```
thebeeland: "Hook that can be used to return extra fields to be displayed for a publish in the details pane."

python/tk_multi_loader/delegate_publish.py
```diff
@@ -108,16 +108,18 @@ class PublishDelegate(shotgun_view.EditSelectedWidgetDelegate):
     how things get rendered.
     """
 
-    def __init__(self, view, action_manager):
+    def __init__(self, view, action_manager, status_model=None):
```
thebeeland: I don't think we should be passing this model around for use by external code like this hook. That's an internal, structural component of the app, and it should stay that way. It looks like you're wanting to get at the convenience method the model object provides to go from a status code to its long name, but that work can be done by the developer in the hook just using the normal API handle and a small amount of logic. I'd prefer that to be their responsibility.

hooks/publish_extra_details.py
```diff
@@ -0,0 +1,24 @@
...
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+from sgtk import Hook
+
+
+class PublishExtraDetails(Hook):
+    """Hook that can be used to add extra fields in the details pane."""
+
+    def execute(self, sg_item, **kwargs):
```
thebeeland: If you're not passing the status model method in (see later comment below) then you can remove the `**kwargs` here. We do use `**kwargs` in some hooks and elsewhere when needed, but if there's not a specific need for it then it's best to stick with a strict signature.

python/tk_multi_loader/delegate_publish_list.py
```diff
@@ -205,6 +205,15 @@ def _format_publish(self, model_index, widget):
         small_text = "<span style='color:#2C93E2'>%s</span> by %s at %s" % (pub_type_str, 
                                                                             author_str,
                                                                             date_str)
+        # if there are extra fields related to published files to be displayed
+        extra_details = sgtk.platform.current_bundle().execute_hook(
+            "publish_extra_details_hook",
+            sg_item=sg_data,
+            get_long_name_method=self._status_model.get_long_name
```
thebeeland: As previously mentioned, I'd prefer we not expose this publicly.

python/tk_multi_loader/dialog.py
```diff
@@ -734,6 +735,15 @@ def __clear_publish_history(pixmap):
                     task_status_str = self._status_model.get_long_name(task_status_code)
                     msg += __make_table_row("Review", task_status_str)
 
+                # if there are extra fields related to published files to be displayed
+                extra_details = sgtk.platform.current_bundle().execute_hook(
+                    "publish_extra_details_hook",
+                    sg_item=sg_item,
+                    get_long_name_method=self._status_model.get_long_name
```
thebeeland: As previously mentioned, I'd prefer we not expose this publicly.

hooks/publish_extra_details.py
000paradox000: I think that **kwargs gives flexibility for implementing any functionality inside the hook

python/tk_multi_loader/delegate_publish.py
000paradox000: How can I access status name using the API inside hook?

barbara-darkshot:
Hey @thebeeland , I've updated the PR code according to your advice. Please let me know if it sounds good to you now :)

jfboismenu:
As discussed in the meeting today, we'll not merge this implementation in favor of a refactor of the custom fields widget used in the Shotgun Panel that offers more flexibility when rendering and would offer the same way of taking over the UI.

PR #80, opened 2019-12-26
Replaced entity type with entity display name

000paradox000:
Currently the delegated widgets are using the entity type. We replaced this value with the entity display name.

image


000paradox000:
Hi @jfboismenu just replaced the method

PR #82, opened 2020-01-27
SG-13278: Python 3 port

jfboismenu:
Another straighforward port. I stumbled on a legacy bug while trying out to refresh the view, but couldn't easily fix it so I backed away.

### DIFF
--- a/python/tk_multi_loader/model_latestpublish.py
+++ b/python/tk_multi_loader/model_latestpublish.py
@@ -326,7 +326,9 @@ class SgLatestPublishModel(ShotgunModel):
         for tree_view_item in self._treeview_folder_items:
 
             # compute and store a hash for the tree view item so that we can access it later
-            tree_view_item_id = id(tree_view_item)
+            # Use the id of the object as the unique identifier, but convert it as a string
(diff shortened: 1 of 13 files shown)


python/tk_multi_loader/model_latestpublish.py
```diff
@@ -326,7 +326,7 @@ def _load_external_data(self):
         for tree_view_item in self._treeview_folder_items:
 
             # compute and store a hash for the tree view item so that we can access it later
-            tree_view_item_hash = str(hash(tree_view_item))
+            tree_view_item_id = id(tree_view_item)
```
jfboismenu: This was interesting. Python 3 does not allow hashing of any data type. Fortunately, a much simpler approach is to take the `id` of the widget.

pscadding: I think this looks fine, as long as this id doesn't get cached anywhere so that it outlives the session. Reading the method description, it seems that it doesn't, so should be good.

jfboismenu:
Yup, dictionary comprehension only works in Python 2.7+.

PR #86, opened 2020-04-07
Loader2 app UI automation

mathurf:
To be able to successfully do ui automation with Loader2 app I needed to add some accessible name.
Otherwise no changes have been done to the app itself.


tests/test_gui.py
```diff
@@ -0,0 +1,349 @@
...
+    def close(self):
+        self.root.buttons["Close"].get().mouseClick()
+
+
+def test_welcome_page(app_dialog):
+    # Validate that Loader app is up
+    assert app_dialog.root.captions["Loader"].exists(), "Cannot find Loader2 app"
+
+    # Validate Welcome page is up
+    if app_dialog.root.floatingwindows["Toolkit Help"].exists() is False:
```
jfboismenu, line 126: I'm assuming this is because the first time the loader is run on a computer we have the welcome screen that shows up by default?

If so, I'd prefer if the `app_dialog` fixture would dismiss the dialog if it shows up. Keep in mind that test_welcome_page is not necessarily the first test that will actually run in your suite, so the pop-up might be present nonetheless. It's better if the app_dialog fixture returns the app in a known consistent state. What do you think?

mathurf: Done :)

PR #89, opened 2020-10-23
[SHOT-3575] Improve Qt Style

staceyoue:
Set style using the QWidget's palette instead of the QApplication or hard coded values to ensure that the style is set correctly according to the widget itself.


eshokrgozar (approved):
Looks good.

python/tk_multi_loader/dialog.py
```diff
@@ -1842,8 +1844,11 @@ def _load_publishes_for_entity_item(self, item):
 
         if show_sub_items:
             # indicate this with a special background color
+            color = self.palette().highlight()
```
eshokrgozar, line 1847: Should this be
```color = self.palette().highlight().color()```?

**Review by eshokrgozar — CHANGES_REQUESTED** [review]
Francis found an issue when showing items in subfolders. 

subfolders

The error is:
```
[DEBUG tk-framework-shotgunutils] User Settings Manager: Retrieving tk-3dsmax:subitems_shown
Traceback (most recent call last):
  File "C:\Users\mathurf\Documents\GitHub\tk-multi-loader2\python\tk_multi_loader\dialog.py", line 1017, in _on_show_subitems_toggled
    self._load_publishes_for_entity_item(item)
  File "C:\Users\mathurf\Documents\GitHub\tk-multi-loader2\python\tk_multi_loader\dialog.py", line 1850, in _load_publishes_for_entity_item
    red=color.red(), green=color().green(), blue=color.blue()
AttributeError: 'PySide2.QtGui.QBrush' object has no attribute 'red'
```
I commented on where I think the issue is, but I haven't tested it.

File: python/tk_multi_loader/dialog.py (hunk 1 above)
Comment (staceyoue) [reply python/tk_multi_loader/dialog.py:1847]: Ah yes, good catch!

File: python/tk_multi_loader/dialog.py [hunk 2]
```diff
@@ -1842,8 +1844,11 @@ def _load_publishes_for_entity_item(self, item):
 
         if show_sub_items:
             # indicate this with a special background color
+            color = self.palette().highlight().color()
             self.ui.publish_view.setStyleSheet(
-                "#publish_view { background-color: rgba(44, 147, 226, 20%); }"
+                "#publish_view { background-color: rgba({red}, {green}, {blue}, 20%); }".format(
+                    red=color.red(), green=color().green(), blue=color.blue()
```
eshokrgozar, line 1850: One more 😄 
```green=color().green()``` should be ```green=color.green()```

**Review by eshokrgozar — CHANGES_REQUESTED** [review]

File: python/tk_multi_loader/dialog.py (hunk 2 above)
Comment (staceyoue) [reply python/tk_multi_loader/dialog.py]: Ah! Got it

File: python/tk_multi_loader/dialog.py [hunk 3]
```diff
@@ -1842,8 +1844,11 @@ def _load_publishes_for_entity_item(self, item):
 
         if show_sub_items:
             # indicate this with a special background color
+            color = self.palette().highlight().color()
             self.ui.publish_view.setStyleSheet(
-                "#publish_view { background-color: rgba(44, 147, 226, 20%); }"
+                "#publish_view { background-color: rgba({red}, {green}, {blue}, 20%); }".format(
```
eshokrgozar, line 1849: Since you're using `format` you'll need to escape/double up the actual braces here:

#publish_view **{{** background-color: rgba({red}, {green}, {blue}, 20%); **}}**

eshokrgozar (changes requested):
There's one more change needed to make this work. Once that's done, here's what it looks like in Nuke:

I'm not sure if the difference in colour is expected or not. I just wanted to point it out:
### Before
old_color
### After
new_color

python/tk_multi_loader/dialog.py
```diff
@@ -1687,9 +1687,11 @@ def _on_search_text_changed(self, pattern, tree_view, proxy_model):
             tree_view.setStyleSheet(
                 """QTreeView { border-width: 3px;
```
eshokrgozar, line 1688: you'll need to escape/double up the actual braces here as well
QTreeView **{{**. ... **}}**

python/tk_multi_loader/dialog.py
staceyoue: Right -- sorry about all these!

python/tk_multi_loader/dialog.py
staceyoue: Fixed

eshokrgozar (approved):

PR #90, opened 2020-11-19
SG-34348 Fixed mutable keyword argument: Multiloader.open_publish

j0yu:
I've noticed a funny behaviour where the `publish_types` keywordsometimes doesn't take effect on the first time `Multiloader.open_publish()` is called.

It was probably from `Multiloader.open_publish()` being called somewhere else previously and having it's mutable kwarg polluted.

See also:
- https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments
- https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

### DIFF
--- a/app.py
+++ b/app.py
@@ -55,7 +55,7 @@ class MultiLoader(sgtk.platform.Application):
         """
         return True
 
-    def open_publish(self, title="Open Publish", action="Open", publish_types=[]):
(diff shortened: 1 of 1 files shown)


carlos-villavicencio-adsk (approved):
Thank you for this contribution ❤️

PR #91, opened 2020-12-17
Adding Azure Agent Name to Shotgun project name

mathurf:
Adding Azure Agent Name to Shotgun project name.
Like that, the Shotgun project will not be overwritten when we will run the UI automation on both, Python 2 and 3

### DIFF
--- a/tests/test_gui.py
+++ b/tests/test_gui.py
@@ -44,7 +44,10 @@ def context():
     )
 
     # Make sure there is not already an automation project created
-    filters = [["name", "is", "Toolkit Loader2 UI Automation"]]
+    project_name = (
+        "Toolkit Panel UI Automation " + os.environ["SHOTGUN_TEST_ENTITY_SUFFIX"]
+    )
+    filters = [["name", "is", project_name]]
     existed_project = sg.find_one("Project", filters)
     if existed_project is not None:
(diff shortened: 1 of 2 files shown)


tests/test_gui.py
```diff
@@ -44,15 +44,18 @@ def context():
     )
 
     # Make sure there is not already an automation project created
-    filters = [["name", "is", "Toolkit Loader2 UI Automation"]]
+    project_name = (
+        "Toolkit Loader2 UI Automation " + os.environ["SHOTGUN_TEST_ENTITY_SUFFIX"]
```
jfboismenu, line 48: This pattern repeats itself in all your repos. Maybe we should refactor it.

in tk-toolchain, could you add a new file called `tk_toolchain/testing.py`. In it, you add a new method, some like 
create_unique_name. Takes a single parameter, the name and then adds the content of the env var if set.

You can look at other files to see how we document methods and tests to see how we turn on/off env vars during a single test.

jfboismenu (changes requested):
Could you refactor that logic of getting the unique name? It would be super useful to simply do
`tk_toolchain.testing.create_unique_name("Toolkit Loader2 UI Automation")`

Don't forget that as written, your test will fail locally if the env var is not set!

jfboismenu (approved):
Everything is awesome!
image

PR #94, opened 2021-05-27
Sg 22646 shotgrid

zamu5:


mathurf (changes requested):
I found some lower case `shotgun` in info.yml
As example...
image

jfboismenu (approved):

mathurf (approved):
Everything looks good

PR #99, opened 2022-03-17
For SG-24449 python3 UI automation

mathurf:
Minor changes for Python 3/PySide2 compatibility

I needed to skip the help screen tests because we are not getting any objectName in Qt and that even if by looking at the help_screen.ui (which is in tk-framework-qtwidgets) the is Qt Object name defines.


000paradox000 (approved):
LGTM

PR #102, opened 2022-12-06
SHOT-4171: Improve filtering

staceyoue:
The original ask here was to be able to specify a default filtering for the Publish File Type.

Took this opportunity to improve the filtering overall in the app. Replaced the publish file type filtering (in bottom left corner) with the Filter Menu (added to top right toolbar). This will provide a more extensive filtering functionality, as well as being consistent with the new filtering added to the ShotGrid Panel and Breakdown2 app.

To specifically address the original request:
- On first time opening the app and opening the Filter Menu, the Publish File Type filter will be shown (no values set)
- User can set the filters they want, then on app close and re-open the filter menu state will be saved/restored

While I was in the Loader2 app, I made a change for ticket [SG-28038](https://jira.autodesk.com/browse/SG-28038) to add a splitter between the main view and the side panels to allow resizing of the panels.

Requires:
- tk-framework-qtwidgets PR:  https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/136 (filter menu improvements/fixes)
- tk-framework-shotgunutils PR: https://github.com/shotgunsoftware/tk-framework-shotgunutils/pull/143 (save/restore splitter state)

**TODO: update minimum version for tk-framework-qtwidgets framework**


NorberMV (changes requested):
I think we should keep the new logos in the resources folder, I've seen some have been replaced by the old ones here
Screenshot 2023-02-16 at 10 33 53 AM

staceyoue:
> I think we should keep the new logos in the resources folder, I've seen some have been replaced by the old ones here Screenshot 2023-02-16 at 10 33 53 AM

Good catch. Not sure how that got reverted to the old logo, but I'll definitely change it back.

NorberMV (approved):
LGTM !

PR #103, opened 2023-01-03
SG-28038: Add splitters to adjust panels and view.

staceyoue:
* Add splitter between left-hand filter panel and main view
* Add splitter between right-hand details panel and main view
* Restore/save splitter state on open/close

Requires tk-framework-shotgunutils changes from PR: https://github.com/shotgunsoftware/tk-framework-shotgunutils/pull/143
- Updated minimum version for tk-framework-shotgunutils due to this dependency -- assuming the next version will be **v5.8.6**


NorberMV (approved):
LGTM!

PR #104, opened 2023-06-22
SG-31007 Loader2 wrong tab/entity selected by default in task context

NorberMV:
This PR fix the issue in which the wrong tab(first tab) in the Loader  app is opened when the Loader is launched in a task context. Now the tab related to the 'entity_type' specified in the tk-multi-loader2.yml is opened.

### DIFF
--- a/python/tk_multi_loader/dialog.py
+++ b/python/tk_multi_loader/dialog.py
@@ -8,7 +8,7 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
-import sys
+
 import sgtk
 from sgtk import TankError
 from sgtk.platform.qt import QtCore, QtGui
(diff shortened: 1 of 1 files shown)


python/tk_multi_loader/dialog.py
```diff
@@ -926,6 +926,20 @@ def _history_navigate_to_item(self, preset, item):
...
+        entity type and entity id.
+
+        :param Sgtk Context ctx: Context object.
+        :param model: The SG model.
+
+        :returns: Model item object or None if not found.
+        """
+        if ctx.task:
+            return model.item_from_entity(ctx.task["type"], ctx.task["id"])
+        return model.item_from_entity(ctx.entity["type"], ctx.entity["id"])
```
carlos-villavicencio-adsk, line 941: Maybe a less verbose approach:

```suggestion
        ctx_object = ctx.task or ctx.entity
        return model.item_from_entity(ctx_object["type"], ctx_object["id"])
```

NorberMV: In this case there will always be a ctx.entity in the context. So, what I'm looking for is that in case there is a Task related to the current context, ctx.task takes precedence over ctx.entity. But I get your point, maybe a ternary operator here would be less verbose.

carlos-villavicencio-adsk (approved):
Nice job!

PR #105, opened 2023-08-16
SHOT-4294: Fix to only show files that match published file types defined in config settings

staceyoue:
* The previous publish type filter widget would only allow showing files that have a published file type that matches the list defined in the config (via `action_mapping`). The new filter menu by default will show all files and allow filtering by any file type
* Add back old published file type model to do the work to extract the published file type filters from the settings
* Add back the proxy filter model logic to filter by the published file types
* Set the published file type filter once on init, this way the view will only get published file types of those defined by the config
* Defaults to show files of all types when config does not have the setting for published types


staceyoue:
Closing, new related PR to come that includes these changes.

PR #108, opened 2023-08-18
SHOT-4297/SHOT-4294: Add config option to use legacy filter widget and fix show only types from config

staceyoue:
There are some minor differences between using the old Published File Type filter widget compared to the new filter menu. If users prefer the old way, they can turn that back on using the new config option. Note that the legacy filter widget cannot be used with the new Filter menu, it is one or the other due to synchronization issues when both filtering is shown.

This also resolves issue for SHOT-4294: when using the new filter menu, only show published file types that are defined via the config.

The code in this PR is almost all strictly adding back functionality from v1.22.1 - no new functionality.

### DIFF
--- a/info.yml
+++ b/info.yml
@@ -24,6 +24,14 @@ configuration:
         default_value: "Loader"
         description: Name to appear on the title of the UI Dialog.
 
+    use_legacy_published_file_type_filter:
+        type: bool
+        default_value: true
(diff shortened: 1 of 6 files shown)


julien-lang (approved):
OK, I see, this is a partial revert of #102.
LGTM comparing it side by side with #102.

PR #109, opened 2023-08-24
SG-31439 Replace help screen with welcome window

minna-feng:


python/tk_multi_loader/__init__.py
```diff
@@ -53,7 +53,6 @@ def show_dialog(app):
     # hide splash screen after loader UI show
     splash.finish(w.window())
 
-    # pop up help screen
+    # pop up welcome window
     if w.is_first_launch():
-        # wait a bit before show window
-        QtCore.QTimer.singleShot(1400, w.show_help_popup)
```
julien-lang, line 59: Why do we remove the timer?

python/tk_multi_loader/dialog.py
```diff
@@ -298,10 +298,6 @@ def __init__(self, action_manager, parent=None):
 
         #################################################
         # set up cog button actions
-        self._help_action = QtGui.QAction("Show Help Screen", self)
-        self._help_action.triggered.connect(self.show_help_popup)
-        self.ui.cog_button.addAction(self._help_action)
```
julien-lang, line 303: Does that mean we are no longer able to display the "help screen" pass the "first time"?

julien-lang (commented):
A couple of comments.

Plus, don't you need to rebuild the resources file when adding/removing PNG files?

python/tk_multi_loader/dialog.py
minna-feng: I thought that was the goal of the ticket. This part removes the button to show it again.

python/tk_multi_loader/__init__.py
minna-feng: It didn't seem to work with the timer, and I wasn't sure it was necessary. But I can try adding it back.

staceyoue: After trying out the branch myself, I don't mind this to show the welcome message without a timer.

staceyoue:
> Plus, don't you need to rebuild the resources file when adding/removing PNG files?

Resources do need to get recompiled, I'm getting an error when trying to run:
image

However, the PNG file change is related to the "Help Screen" that is shown when the "Show items in subfolders" is checked. Do we also just want to remove this help screen pop up altogether, and make sure documentation includes this functionality?

staceyoue (commented):
Just a couple things:

1. The link to the documentation from the `Welcome` dialog goes [here](https://developer.shotgridsoftware.com/d587be80/?title=Integrations+User+Guide), while the `View Documentation` link goes [here](https://help.autodesk.com/view/SGSUB/ENU/?guid=SG_Supervisor_Artist_sa_integrations_sa_integrations_user_guide_html).

My thoughts are that the links should go to the same place, and should point directly to the Loader App docs. Also with the new documentation work coming soon, we may point there instead now.

2. This is how the welcome message looks when I run with VRED:

image

Just wondering if this new dialog has gone through UX/Design? A few things come to mind:
* The dialog window title should be set to the loader app name, not the DCC (e.g. VRED)
* The hyperlink color (I think) should match the Qt Palette `Highlight` color
* For the text, I would suggest removing the bold font and changing the message to something like:  "Welcome! Learn more about the Loader App here."

But this is just my opinion :)

python/tk_multi_loader/dialog.py
```diff
@@ -352,6 +348,26 @@ def __init__(self, action_manager, parent=None):
         show_details = self._settings_manager.retrieve("show_details", False)
         self._set_details_pane_visiblity(show_details)
 
+        ################################################
+        # set up welcome window
+        welcome_msg = (
+            "<strong>Welcome - click <a "
+            "href='https://developer.shotgridsoftware.com/d587be80/?title=Integrations+User+Guide' "
```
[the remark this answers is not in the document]
NorberMV: Sure I'll confirm the right link with Shayna, and I'll make sure that the 'View Documentation' link points there as well.

shaynacohen (approved):
Docs link looks perfect!

python/tk_multi_loader/dialog.py
```diff
@@ -352,6 +348,30 @@ def __init__(self, action_manager, parent=None):
...
+
+        self.welcome_widget = QtGui.QMessageBox(
+            QtGui.QMessageBox.Information,
+            "",
+            welcome_msg,
+            QtGui.QMessageBox.Ok,
+        )
+        # force the QMessageBox to be on top of other dialogs.
+        self.welcome_widget.setWindowFlags(
+            self.welcome_widget.windowFlags() | QtCore.Qt.WindowStaysOnTopHint
```
julien-lang, line 369: Could you add a comment explaining why you do that instead of using the parent parameter, please?

For the record, this is similar to shotgunsoftware/tk-core#899 and shotgunsoftware/tk-core#907.

I regret the choice I made in tk-core. I would revert that and move the QMessageBox creation in the function that needs it (instead of __init__) so I can use the parent parameter and not have to deal with window flags.

python/tk_multi_loader/dialog.py
```diff
@@ -352,6 +348,30 @@ def __init__(self, action_manager, parent=None):
         show_details = self._settings_manager.retrieve("show_details", False)
         self._set_details_pane_visiblity(show_details)
 
+        ################################################
+        # set up welcome window
+        highlight_color = self.palette().highlight().color().name()
+        welcome_msg = (
+            "Welcome! Learn more about the Loader App <a "
+            "href='https://help.autodesk.com/view/SGDEV/ENU/?guid=SG_Supervisor_Artist_sa_integrations_sa_integrations_user_guide_html#the-loader' "
```
julien-lang, line 356: Why not use the documentation_url variable here instead?

python/tk_multi_loader/dialog.py
```diff
@@ -352,6 +348,30 @@ def __init__(self, action_manager, parent=None):
...
+            "style='text-decoration: none;'><font color='%s'>here.</font></a>"
```
julien-lang, line 357: Can you move the dot out of the link please?

python/tk_multi_loader/dialog.py
```diff
@@ -352,6 +348,30 @@ def __init__(self, action_manager, parent=None):
...
+            % highlight_color
```
julien-lang, line 358: Can we use the format function instead? I find it easier to visualize variable on long strings.
```python
welcome_msg = (
    "Welcome! Learn more about the Loader App <a href='{url}' "
    "style='text-decoration: none;'><font color='{color}'>here</font></a>."
    "".format(
        color=highlight_color,
        url=documentation_url
    )
)
```

julien-lang (commented):
What is the empty resources_rc.py file about?

NorberMV:
> What is the empty resources_rc.py file about?
Oh, It was probably generated when I was testing something with the build_resources.sh file, I'll remove it

python/tk_multi_loader/dialog.py
NorberMV: Ok, great I'll try that.

julien-lang (approved):
LGTM!

PR #113, opened 2023-11-16
Add has_resources flag, enabling resources validation.

NorberMV:


carlos-villavicencio-adsk (commented):
Should this PR also need to remove the `mogrify` from build_resources?

julien-lang (commented):
Can you update the `resources/build_resources.sh` script to remove the mogrify command please?

NorberMV:
> Should this PR also need to remove the `mogrify` from build_resources?
Sure, I believe once this https://github.com/shotgunsoftware/tk-ci-tools/pull/46 PR is approved, we can begin implementing this and remove the "mogrify" command from some of the bash scripts. 
The idea is to validate the profiles of the PNG files in CI. If there are any issues, you can use an image editor or image processing tool that supports color profile manipulation (like ImageMagick's "mogrify" command) on those specific files.

PR #122, opened 2024-07-04
SG-35762 first phase python2 removing

eduardoChaucaGallegos:
- Six str types replaced by native str python type.
- Six ensure functions replaced with function from sgutils


carlos-villavicencio-adsk (commented):
We can use `str` instead I believe.

carlos-villavicencio-adsk (approved):

julien-lang (approved):

PR #137, opened 2026-05-22
SG-43456 [Phases 1-2] Enable FlowAM integration

carlos-villavicencio-adsk:
Base branch for tk-multi-loader2 v2.0.0

## Approved Contents

- MEDM Abstractions
- #138 
- #139 
- #140 
- #142 
- #143
- #144 
- #145
- #147 
- #148

ℹ️ This should not affect or break the loader functionality.

## Dependencies

- tk-core:  https://github.com/shotgunsoftware/tk-core/pull/1108
- tk-maya:  https://github.com/shotgunsoftware/tk-maya/pull/130
- tk-houdini:  https://github.com/shotgunsoftware/tk-houdini/pull/103
- tk-nuke:  https://github.com/shotgunsoftware/tk-nuke/pull/134
- tk-desktop:  https://github.com/shotgunsoftware/tk-desktop/pull/224

## Config Requirements

- core/core_api.yml
    - Use a descriptor to `master`

## TODO

- [x] Remove the `tk-framework-flowam` entry on `info.yml` that is causing CI to fail #145 
- [x] Add flowam config instructions
- [x] Tested on Maya
- [x] Tested on Desktop
- [x] Remove `tk_core_ref` from `azure-pipelines.yml` after https://github.com/shotgunsoftware/tk-core/pull/1108 is merged

PR #138, opened 2026-05-26
SG-43457 [Part2] Migration of MEDM main logic and resources

carlos-villavicencio-adsk:
Base branch: #137

## Loader core logic
This is where most of the changes were made for displaying information and UI changes.

- dialog
- utils
- delegates (publish list and publish thumb)
- manager
- info.yml settings / constants / display info for configurable items 
- Other UI/UX updates.


python/tk_multi_loader/dialog.py
```diff
@@ -785,158 +1059,139 @@ def __clear_publish_history(pixmap):
...
+                    self._publish_history_proxy.setSourceModel(self._medm_history_model)
+                    self._medm_history_model.load_data(sg_data)
+                else:
+                    self._publish_history_proxy.setSourceModel(
+                        self._publish_history_model
+                    )
+                    self._publish_history_model.load_data(sg_data)
+                self._update_history_view_height()
+
+            self.ui.details_header.updateGeometry()
```
chenm1adsk, line 1198: There is a repeat line of `self.ui.details_header.updateGeometry()` in line 1196, is this a typo?

python/tk_multi_loader/delegate_publish_list.py
```diff
@@ -144,62 +215,26 @@ def _format_publish(self, model_index, widget):
...
-
-        # Publish Name Version 002
         sg_data = shotgun_model.get_sg_data(model_index)
+        entity_type = sg_data.get("type") if sg_data else None
+
+        main_text_fields = ["name", "version_number", "entity", "task"]
+        default_small_text_fields = ["created_by", "created_at"]
+
+        configured_fields = self._list_entity_fields.get(entity_type, [])
+        filter_fields = list(set(main_text_fields + default_small_text_fields))
```
chenm1adsk: let‘s use `  filter_fields = list(dict.fromkeys(main_text_fields + default_small_text_fields))` here instead

python/tk_multi_loader/delegate_publish_list.py
```diff
@@ -144,62 +215,26 @@ def _format_publish(self, model_index, widget):
...
+
         main_text = "<b>%s</b>" % (sg_data.get("name") or "Unnamed")
 
         version = sg_data.get("version_number")
-        vers_str = "%03d" % version if version is not None else "N/A"
+        if version == MedmLatestPublishModel.DRAFT_VERSION_IDENTIFIER:
```
chenm1adsk: Should we put this DRAFT_VERSION_IDENTIFIER in constants.py?

python/tk_multi_loader/api/manager.py
```diff
@@ -86,16 +86,22 @@ def get_actions_for_publish(self, sg_data, ui_area):
             # this publish does not have a type
             publish_type = "undefined"
         else:
-            publish_type = publish_type_dict["name"]
+            publish_type = (
```
chenm1adsk, line 91: Can we have a line of comment explain why check code first and then check name if code doesn't exists?

python/tk_multi_loader/api/manager.py
```diff
@@ -86,16 +86,22 @@ def get_actions_for_publish(self, sg_data, ui_area):
...
 
         # check if we have logic configured to handle this publish type.
         mappings = self._bundle.get_setting("action_mappings")
         if not mappings:
             return []
         # returns a structure on the form
         # { "Maya Scene": ["reference", "import"] }
-        actions = mappings.get(publish_type, [])
-        actions.extend(mappings.get("All", []))
+        # IMPORTANT: Make a copy to avoid modifying the cached config dict
```
chenm1adsk, line 103: I made this change for poc config specifically. Can we confirm if the original behavior is expected or it is a bug? Because this line of change will require user to change their config to list all actions for each publish_type they defined. The action in `All` will only append when the publish_type is not defined. I am not sure if this is something we want to add to loader.
cc: @julien-lang

python/tk_multi_loader/model_entity.py
```diff
@@ -75,8 +77,42 @@ def __init__(self, parent, entity_type, filters, hierarchy, bg_task_manager):
...
+        )
+        constant_middle_panel_fields = ENTITY_TYPE_MIDDLE_PANEL_FIELDS.get(
+            entity_type, []
+        )
+        constant_detail_panel_fields = ENTITY_TYPE_DETAIL_PANEL_FIELDS.get(
+            entity_type, []
+        )
+
+        all_fields = list(
+            set(
```
chenm1adsk: Let's use `dict.fromkeys` to replace `set`

python/tk_multi_loader/model_latestpublish.py
```diff
@@ -270,12 +270,33 @@ def _do_load_data(self, sg_filters, treeview_folder_items):
...
+        )
+
         if publish_entity_type == "PublishedFile":
             self._publish_type_field = "published_file_type"
         else:
             self._publish_type_field = "tank_type"
 
-        publish_fields = [self._publish_type_field] + constants.PUBLISHED_FILES_FIELDS
+        publish_fields = [self._publish_type_field] + list(
+            set(
```
chenm1adsk, line 293: I am not sure why I didn't replace all place to use `dict.fromkeys` instead of `set`. Can you double check?

python/tk_multi_loader/model_publishhistory.py
```diff
@@ -58,13 +59,26 @@ def load_data(self, sg_data):
...
+
         if publish_entity_type == "PublishedFile":
             publish_type_field = "published_file_type"
         else:
             publish_type_field = "tank_type"
 
         # fields to pull down
-        fields = [publish_type_field] + constants.PUBLISHED_FILES_FIELDS
+        fields = [publish_type_field] + list(
+            set(constant_detail_panel_fields + detail_fields + flow_am_internal_fields)
```
chenm1adsk: I am not sure why I didn't replace all place to use dict.fromkeys instead of set. Can you double check?

chenm1adsk (commented):
The remaining looks good.

python/tk_multi_loader/ui/widget_publish_history.py
```diff
@@ -23,11 +23,11 @@ class Ui_PublishHistoryWidget(object):
     def setupUi(self, PublishHistoryWidget):
         if not PublishHistoryWidget.objectName():
             PublishHistoryWidget.setObjectName(u"PublishHistoryWidget")
-        PublishHistoryWidget.resize(1226, 782)
+        PublishHistoryWidget.resize(1226, 130)
```
julien-lang, line 26: What is this change about? How is not going to create a regression?

python/tk_multi_loader/constants.py
```diff
@@ -13,26 +13,88 @@
...
+    "Shot": [
+        "name",
+        "description",
+        "sg_status_list",
+        "project",
+        "sg_sequence",
+        "sg_cut_in",
+        "sg_cut_out",
+    ],
+}
```
julien-lang, line 97: Why do we need all these new things?

python/tk_multi_loader/ui/widget_publish_history.py
julien-lang: Same comment for all the UI changes in this PR

python/tk_multi_loader/__init__.py
```diff
@@ -8,17 +8,71 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
-from .api import LoaderManager
-from .open_publish_form import open_publish_browser
+from .api import LoaderManager  # noqa: F401
+from .open_publish_form import open_publish_browser  # noqa: F401
```
julien-lang, line 12: why do we need this change?

python/tk_multi_loader/__init__.py
```diff
@@ -8,17 +8,71 @@
...
+
+    The extra flags (SWP_NOMOVE, SWP_NOSIZE, SWP_NOACTIVATE) prevent accidental
+    window movement or resizing during the operation.
+    """
+    if win is None or not win.isWidgetType():
+        return
+
+    if sys.platform == "win32":
+        try:
+            import ctypes
```
julien-lang, line 45: I think this needs to be in a dedicated PR with a dedicated Jira ticket.
And also, the code should go in tk-nuke.

python/tk_multi_loader/constants.py
chenm1adsk: This was initially coming from a request from Kevin to make the loader fields configurable.
Maybe we want to revisit this part of the code

python/tk_multi_loader/ui/widget_publish_history.py
carlos-villavicencio-adsk: See FFACE-2452

python/tk_multi_loader/constants.py
carlos-villavicencio-adsk: See FFACE-2461.

I already discussed this with product and he's ok including this to the upstream.

python/tk_multi_loader/__init__.py
carlos-villavicencio-adsk: I'm importing these somewhere else, and it's easier to keep all imports consistent from a single source.

python/tk_multi_loader/__init__.py
carlos-villavicencio-adsk: We had a dedicated ticket. See FFACE-2832.

Also, this fix doesn't affect Nuke only, as far as I remember. It fixes desktop loader as well (See FFACE-3085).

python/tk_multi_loader/delegate_publish_list.py
carlos-villavicencio-adsk: Great suggestion!

python/tk_multi_loader/api/manager.py
carlos-villavicencio-adsk: I didn't understand this comment. Let's catch up later.

carlos-villavicencio-adsk: To be confirmed.

PR #140, opened 2026-06-02
SG-43460 [Part4] Build dialogs and template utils

carlos-villavicencio-adsk:
Base branch #139

# Build dialogs

Migrate the code in the two custom UIs

- Build Asset Dialog
- Build Template Dialog

And related utilities into tk-multi-loader2.


python/tk_multi_loader/build_asset_dialog.py
```diff
@@ -0,0 +1,214 @@
...
+        Args:
+            text (str): The new build option selected.
+        """
+        is_template_mode = self._CreateMode(text) == self._CreateMode.TEMPLATE
+
+        self.setUpdatesEnabled(False)
+        self.ui.templateWidget.setVisible(is_template_mode)
+        self.ui.pipeline_step_combo_box.clear()
+
+        if is_template_mode and self.step:
```
chenm1adsk, line 152: We should call `self.get_pipeline_steps()` here as well, otherwise `self.pipeline_steps` is always empty and will cause issue when later we try to ` pipeline_step = self.pipeline_steps[step]`  in `get_pipeline_step_templates()`

python/tk_multi_loader/build_template_dialog.py
```diff
@@ -0,0 +1,136 @@
...
+        )
+
+        self.ui.pipeline_step_combo_box.addItems(pipeline_steps)
+
+        # Obtain the button to disconnect the default slot to avoid to close the
+        # dialog if the validation fails
+        self.ok_button = self.ui.build_template_button_box.button(
+            QtGui.QDialogButtonBox.Ok
+        )
+        self.ok_button.clicked.disconnect()
```
chenm1adsk, line 69: Claude give me this suggestion but I am not sure if it is correct, could you please double check?

> ok_button.clicked has no Python-side slots connected, so disconnect() raises   
>   RuntimeError. The internal Qt connection from ok_button.clicked to             
>   button_box.accepted() is managed in C++ and invisible to Python. Fix:        
>   disconnect button_box.accepted instead:                                        
>                                                                                  
>   `self.ui.build_template_button_box.accepted.disconnect() `

python/tk_multi_loader/build_template_dialog.py
```diff
@@ -0,0 +1,136 @@
...
+            self.project = _Project(project_id)
+        except _FlowError as exc:
+            raise ValueError(f"Project not found: {project_id}") from exc
+
+        if not pipeline_steps:
+            raise ValueError("Pipeline steps must be provided to populate the dialog.")
+
+        self.step = None
+        self.template = None
+        self.description = None
```
chenm1adsk, line 52: Did we miss initialize self.mode = None here as well?

chenm1adsk (commented):
The remaining looks good to me.

python/tk_multi_loader/build_template_dialog.py
carlos-villavicencio-adsk: I don't think we should worry about this.

python/tk_multi_loader/build_asset_dialog.py
carlos-villavicencio-adsk: I don't  think this is something we should worry about. I've been logging this scenario, and `self.step` is mostly `None` all the time, and `self.pipeline_steps` can be `{}`.

PR #141, opened 2026-06-08
feat: [SG-43418] Replace env var side channel with show_dialog parameters

chenm1adsk:
The Loader previously used TK_FLOWAM_REVISION_ID_* environment variables
to pass am_revision_id to the Publisher, which caused race conditions when
multiple Publisher dialogs were open simultaneously.

Replace with clean parameter passing via show_dialog(context, root_item_properties):
- When publishing new generic asset: pass entity_ctx so Publisher can pre-fill the context widget and use task/entity for MEDM hierarchy placement.
- When publishing existing generic asset revision: pass context=None and root_item_properties={"am_revision_id": revision_id} so the publish hook receives the correct asset identifier without env vars or shared state.


carlos-villavicencio-adsk (approved):
LGTM

PR #142, opened 2026-06-09
SG-43626 Fix bug when details panel collapses/expands

carlos-villavicencio-adsk:
Base branch: #140

The bug: `DeleteWhenStopped` makes Qt destroy the C++ animation object when it finishes, but `self._current_animation` still holds the dead Python wrapper. On the next toggle, .state() crashes on the deleted object.

Fix: connect finished to clear `self._current_animation`, so the guard if `self._current_animation` becomes False after it's gone.

When the animation finishes, Qt deletes the C++ object and the finished signal fires — clearing self._current_animation to None before anything can touch the dead wrapper again.


chenm1adsk (approved):
LGTM!

PR #143, opened 2026-06-09
SG-43627 Fix empty space gap on the details panel

carlos-villavicencio-adsk:
Base branch: #142

**Empty gap below versions list** - history_view had a hardcoded minimumHeight: 350 in the UI file, forcing it to always occupy 350px regardless of content. Removed this constraint and switched the size policy to Preferred so the existing dynamic sizing code (_patch_history_view_sizing / _update_history_view_height) could actually take effect. _update_history_view_height was also updated to call setMaximumHeight directly rather than relying on updateGeometry alone, which Qt treats as advisory.

The change only affects the .ui file, but the rest of the files are generated by build resources script.


chenm1adsk (approved):
Lgtm! but please remember resolve the failing checks :)

PR #144, opened 2026-06-10
SG-43677 Fix hooks to support multiple keyword arguments

carlos-villavicencio-adsk:
Base branch: #137 

DCCs that are not supported with FlowAM integration crashes their hooks due to unexpected arguments. This PRs addresses this issue by just adding `**kwargs` and silently ignoring them when not needed.

### DIFF
--- a/hooks/tk-3dsmaxplus_actions.py
+++ b/hooks/tk-3dsmaxplus_actions.py
@@ -24,7 +24,7 @@ class MaxActions(HookBaseClass):
     ##############################################################################################################
     # public interface - to be overridden by deriving classes
 
-    def generate_actions(self, sg_publish_data, actions, ui_area):
+    def generate_actions(self, sg_publish_data, actions, ui_area, **kwargs):
         """
         Returns a list of action instances for a particular publish.
         This method is called each time a user clicks a publish somewhere in the UI.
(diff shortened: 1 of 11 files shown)


chenm1adsk (approved):
LGTM!

hooks/tk-3dsmaxplus_actions.py
julien-lang, line 1: Should we remove this file? I'm surprised we missed it!

hooks/tk-3dsmaxplus_actions.py
```diff
@@ -24,7 +24,7 @@ class MaxActions(HookBaseClass):
     ##############################################################################################################
     # public interface - to be overridden by deriving classes
 
-    def generate_actions(self, sg_publish_data, actions, ui_area):
+    def generate_actions(self, sg_publish_data, actions, ui_area, **kwargs):
```
julien-lang, line 27: Why now?

carlos-villavicencio-adsk: Because I'm injecting a new argument to the hook for flowam integration to work.

This is not needed on all DCCs (i.e. 3ds, photoshop) but it crashes due to the unexpected argument. So this is just adding the placeholder. 

You can see the difference on this file and compare it to tk-maya_actions.py to have a better picture.

carlos-villavicencio-adsk:
Merging this. Any further updates can be done against #137

PR #145, opened 2026-06-15
SG-43459 Migrate Build scene/template workflows

carlos-villavicencio-adsk:
Base branch #137 

## Requirements

- tk-core:  https://github.com/shotgunsoftware/tk-core/pull/1108
- tk-maya:  https://github.com/shotgunsoftware/tk-maya/pull/130
- tk-houdini:  https://github.com/shotgunsoftware/tk-houdini/pull/103
- tk-nuke:  https://github.com/shotgunsoftware/tk-nuke/pull/134
- tk-desktop:  https://github.com/shotgunsoftware/tk-desktop/pull/224

## Description

This will prevent having these workflows dependant from tk-framework-flowam.

## TODO

- [x] Remove references for framework-flowam on `flowam_actions.py`
- [x] Remove references for framework-flowam on entity, latest publish, and publish history models.
- [x] Remove requirement on `info.yml`
- [x] Tested on Maya
- [x] Tested on Desktop
- [ ] Remove `tk_core_ref` from `azure-pipelines.yml`


hooks/tk-houdini_actions.py
```diff
@@ -102,22 +103,16 @@ def generate_actions(self, sg_publish_data, actions, ui_area, **kwargs):
...
-                    "FlowAM is enabled but no Asset Management base object was passed to the action hook. "
-                    "FlowAM specific actions will not be generated."
-                )
+        if self.parent.context.flow_project_id:
+            flowam_actions = app.flowam.FlowAMActions()
 
             if "open" in actions and sg_publish_data.get("type") == "PublishedFile":
                 if (
-                    am_base_obj._is_local_draft(sg_publish_data)
+                    is_local_draft(sg_publish_data.get("sg_flow_revision_id"))
```
yungsiow: Surprised this logic is still in here?  Is this still catering to PublishedFile proxies?

python/tk_multi_loader/api/manager.py
chenm1adsk, line 106: Not introduced by this PR but as I mentioned in phase1 PR, this will change the behaviour of when to append  actions in "All", we shouldn't introduce to tk-multi-loader2.

python/tk_multi_loader/flowam/flowam_actions.py
```diff
@@ -366,21 +323,17 @@ def _on_build_template_dialog_accepted(
...
+        create_inputs = CreateTemplateInputs(
             sg_pipeline_step=dialog.step,
             am_project_id=sg_flow_am_id,
             template_name=dialog.template,
             create_mode=dialog.mode,
         )
-        draft_info = flow_module.asset_management.create_template_workfile(
-            create_inputs
-        )
+        draft_info = create_template_workfile(create_inputs)
```
chenm1adsk: Should we add handle exception like `create_dcc_workfile`?
```python
try:
    asset = create_dcc_workfile(create_inputs)
    ...
except exceptions.CreateAssetError as exc:
    self._app.log_error(...)
    QtGui.QMessageBox.critical(parent_window, "Error", str(exc))
    return
```

python/tk_multi_loader/flowam/reference.py
```diff
@@ -0,0 +1,139 @@
...
+        msg = "Revision does not have a source component to be referenced."
+        raise CreateReferenceError(input_id=revision_id, details=msg)
+    if not os.path.exists(file_path):
+        msg = f"Source file does not exist in storage: {file_path}. "
+        msg += "Fetching the revision was not successful!"
+        raise CreateReferenceError(input_id=revision_id, details=msg)
+
+    # Create reference
+    depdata = engine.flow_host.create_reference(file_path, namespace=revision.name)
+    return depdata.file_path
```
chenm1adsk, line 89: Compared with poc framework, are we dropping sequence file support for both reference and create reference link?

python/tk_multi_loader/flowam/__init__.py
```diff
@@ -8,13 +8,11 @@
...
-is enabled in the app configuration.  All models share a single
-:class:`~medm.shared_cache.MedmSharedCache` and
-:class:`~medm.thumbnail_service.MedmThumbnailService` instance injected by
-the dialog at construction time.
+This package provides drop-in replacements for the standard Shotgun-based
+Loader models and actions, backed by Flow Asset Management (FlowAM) instead
+of the ShotGrid REST API.
 """
 
 from .entity_model import MedmEntityModel
```
chenm1adsk, line 18: Are we going to rename models to have "Flow" prefix instead of "Medm" prefix?

python/tk_multi_loader/flowam/create.py
```diff
@@ -0,0 +1,489 @@
...
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""
+This module contains utilities for creating medm assets. Some of these functions
```
chenm1adsk: Maybe update the docstring to "creating flow assets" instead.

hooks/tk-desktop_actions.py
chenm1adsk: Not in this PR but it seems this log is missing arguments and also we should use f-string style log.

hooks/tk-houdini_actions.py
```diff
@@ -249,27 +244,26 @@ def execute_action(self, name, params, sg_publish_data, **kwargs):
         # -----------------------
         # FlowAM specific actions
         # -----------------------
-        enable_flowam = app.get_setting("enable_flowam", False)
-        if enable_flowam:
-            am_base_obj = kwargs.get("am_base_obj")
+        if self.parent.context.flow_project_id:
```
chenm1adsk, line 249: I feel like it is better to just create a function to group flow logic.
For example create a _flow_execute_action and called it when flow project id exists

hooks/tk-maya_actions.py
```diff
@@ -129,24 +130,18 @@ def generate_actions(self, sg_publish_data, actions, ui_area, **kwargs):
...
         # -----------------------
-        enable_flowam = app.get_setting("enable_flowam", False)
-        if enable_flowam:
-            am_base_obj = kwargs.get("am_base_obj")
-            if not am_base_obj:
-                raise Exception(
-                    "FlowAM is enabled but no Asset Management base object was passed to the action hook. "
-                    "FlowAM specific actions will not be generated."
-                )
+        if self.parent.context.flow_project_id:
```
chenm1adsk, line 133: same for maya and nuke actions

python/tk_multi_loader/flowam/latestpublish_model.py
```diff
@@ -17,23 +17,18 @@
...
+from .utils import build_draft_sg_dict
 from .utils import is_structural_asset as _is_structural_asset_util
-
-if TYPE_CHECKING:
-    from flow.data import Asset
-    from flow.sandbox import DraftInfo
-else:
-    Asset = Any
-    DraftInfo = Any
+from .utils import resolve_publish_type
```
chenm1adsk, line 31: Wondering why we don't combine 29-31 together?

chenm1adsk (commented):
Left some question of improvement that I am not sure if we should aim for in the scope of this ticket.

python/tk_multi_loader/flowam/__init__.py
carlos-villavicencio-adsk: In this case, I believe Medm is more accurate because we're refering to data models.

python/tk_multi_loader/flowam/reference.py
carlos-villavicencio-adsk: Can you elaborate more on this?

hooks/tk-desktop_actions.py
carlos-villavicencio-adsk: These logs shouldn't be there.Good catch.

hooks/tk-maya_actions.py
carlos-villavicencio-adsk: This is legacy code. Let's not worry about this now.

python/tk_multi_loader/flowam/latestpublish_model.py
carlos-villavicencio-adsk: This was done automatically by `isort` and `black`

python/tk_multi_loader/flowam/file.py
```diff
@@ -0,0 +1,183 @@
...
+
+class InvalidDraftError(exceptions.FlowError):
+    def __init__(self, *args, draft_id: str, **kwargs):
+        """
+        Args:
+            draft_id: Id that uniquely identifies a draft in local sandbox.
+        """
+        message = f'Draft id "{draft_id}" is invalid.'
+        super().__init__(message, *args, **kwargs)
+        self.draft_id = draft_id
```
yungsiow: This error is already available in the flow_integration_sdk.exceptions module.

python/tk_multi_loader/flowam/flowam_actions.py
```diff
@@ -130,7 +117,7 @@ def _build_new_scene(self, sg_publish_data: dict) -> None:
         :param sg_publish_data: Shotgun data dictionary with all the standard publish fields.
         """
         parent_window = self._get_dialog_parent()
-        sg_flow_am_id = self._app.context.project.get("sg_flow_am_id")
+        sg_flow_am_id = self._get_flowam_id()
```
yungsiow: I still find this variable name confusing.  Is it possible to get rid of the sg_ prefix?  Or are there still reasons we need to keep it in?

python/tk_multi_loader/flowam/flowam_actions.py
```diff
@@ -283,9 +242,7 @@ def _discard_draft(self, sg_publish_data: dict) -> None:
         )
 
         if message_response == QtGui.QMessageBox.StandardButton.Yes:
-            flow_module.asset_management.discard_draft(
-                sg_publish_data.get("sg_flow_revision_id")
-            )
+            sandbox.discard_draft(sg_publish_data.get("sg_flow_revision_id"))
```
yungsiow: So actually replacing the original asset_management module function with the lower level sandbox.discard_draft() function here isn't enough.  In the original asset_management function, we also did some work to clear the current scene if the draft was currently open.  Please take a look a be sure to port that functionality over as well, thanks!  (it's just a few extra lines of code).

python/tk_multi_loader/flowam/reference.py
```diff
@@ -0,0 +1,139 @@
...
+from __future__ import annotations  # needed for Houdini support
+
+import os
+
+import sgtk
+from tank_vendor.flow_integration_sdk import globals, exceptions, objects
+
+
+class CreateReferenceError(exceptions.FlowError):
+    def __init__(self, *args, input_id: str = "", file_path: str = "", **kwargs):
```
yungsiow: I would recommend simplifying this error class.  Before, it was being shared by both these higher-level functions and the host-level reference functions, which would pass in a file_path.  Since the host-level functions are no longer using this error, you can remove the file_path parameter.  You can even make input_id mandatory since you're always passing it it.

python/tk_multi_loader/flowam/reference.py
```diff
@@ -0,0 +1,139 @@
...
+    file_path = revision.get_storage_component_path(
+        component_purpose=globals.SOURCE_PURPOSE
+    )
+    if file_path is None:
+        msg = "Revision does not have a source component to be referenced."
+        raise CreateReferenceError(input_id=revision_id, details=msg)
+    if not os.path.exists(file_path):
+        msg = f"Source file does not exist in storage: {file_path}. "
+        msg += "Fetching the revision was not successful!"
+        raise CreateReferenceError(input_id=revision_id, details=msg)
```
yungsiow, line 80: Hi, I think Ming may have also mentioned this, but this is missing the file sequence support in the original utility.  Can we put that back in please?

python/tk_multi_loader/flowam/reference.py
```diff
@@ -0,0 +1,139 @@
...
+    # Get path to source path of revision in local storage
+    file_path = revision.get_storage_component_path(
+        component_purpose=globals.SOURCE_PURPOSE
+    )
+    if file_path is None:
+        msg = "Revision does not have a source component to be referenced."
+        raise CreateReferenceError(input_id=revision_id, details=msg)
+    if not os.path.exists(file_path):
+        msg = f"Source file does not exist in storage: {file_path}"
+        raise CreateReferenceError(input_id=revision_id, details=msg)
```
yungsiow, line 139: Same here, let's make sure the file sequence support is preserved.

python/tk_multi_loader/flowam/reference.py
carlos-villavicencio-adsk: I didn't understand this. Can we take a look together, please?

python/tk_multi_loader/flowam/reference.py
carlos-villavicencio-adsk: ~Done in commit 71eedd485a3df4b35f2a3a4e5744671123a74d6f, can you take a look to sign off?~

python/tk_multi_loader/flowam/flowam_actions.py
yungsiow: Looks good now, thanks!

python/tk_multi_loader/flowam/reference.py
yungsiow: Sorry I think you may have posted to the wrong comment?  This commit seems to only deal with the discard draft functionality.  And yes, that looks good!

python/tk_multi_loader/flowam/reference.py
carlos-villavicencio-adsk: Fixed in 113d9464ce7dd01b6fc94e02e4c09f8bc89346b9

python/tk_multi_loader/flowam/reference.py
carlos-villavicencio-adsk: Fixed in 113d9464ce7dd01b6fc94e02e4c09f8bc89346b9

python/tk_multi_loader/flowam/reference.py
carlos-villavicencio-adsk: Fixed in b74125e269ddec007d0a42d2a0fd2755c1e4d295

python/tk_multi_loader/flowam/reference.py
```diff
@@ -79,10 +79,18 @@ def reference_revision(revision_id: str) -> str:
     if file_path is None:
         msg = "Revision does not have a source component to be referenced."
         raise CreateReferenceError(input_id=revision_id, details=msg)
+    file_seq_comp = revision.find_component(
+        type_id=schema.get_schema_id(globals.FILE_SEQ_TYPE)
+    )
     if not os.path.exists(file_path):
```
yungsiow: This line should be :  `if not file_seq_comp and not os.path.exists(file_path):`

python/tk_multi_loader/flowam/reference.py
```diff
@@ -17,21 +17,17 @@
...
 
 class CreateReferenceError(exceptions.FlowError):
-    def __init__(self, *args, input_id: str = "", file_path: str = "", **kwargs):
+    def __init__(self, *args, input_id: str, **kwargs):
         """
         Args:
             input_id: Id of revision or version being referenced.
-            file_path: File being referenced.
         """
         if input_id:
```
yungsiow: Don't think we need this if now, because input_id should always be provided.

yungsiow (approved):
Looks good!  Thanks Charlie!

PR #147, opened 2026-06-22
feat: [SG-43419] Launch publisher via engine command with single_file_mode support

chenm1adsk:
Summary
----
Replace direct app lookup + non-existent _set_context() call with the established engine command callback pattern. 
Grab the publisher app from the matched command's properties to call show_dialog() directly, passing single_file_mode=True when the action is a Flow AM republish ("publish"), so the dialog restricts users to dropping or browsing a single file.

Tests
----
Verified when single_file_mode turned on, only accept one file in publish dialog
Screenshot 2026-06-22 at 3 11 38 PM
Also verified that this restriction doesn't cause regression when single_file_mode is off


carlos-villavicencio-adsk (approved):
LGTM

PR #148, opened 2026-07-07
Remove config dependency for entity mapping

carlos-villavicencio-adsk:
Base PR #137 

We used to have multiple mappings per DCC in the testing config. For now, let's move them to the engine hooks and don't expose them to the yml settings.

### DIFF
--- a/hooks/tk-houdini_actions.py
+++ b/hooks/tk-houdini_actions.py
@@ -281,6 +281,36 @@ class HoudiniActions(HookBaseClass):
         if name == "file_cop":
             self._file_cop(path, sg_publish_data)
 
+    def action_mappings(self) -> dict:
+        """
+        Returns the action mappings for the loader app.
+
+        :returns: Dictionary of action mappings.
+        """
+        app = self.parent
+        if app.flowam_available:
+            return {
+                "All": ["reference_copy_link"],
+                "Houdini Workfile": ["open", "discard_draft"],
+            }
+
+        return {}
+
+    def entity_mappings(self) -> dict:
+        """
+        Returns the entity mappings for the loader app.
+
+        :returns: Dictionary of entity mappings.
+        """
+        app = self.parent
+        if app.flowam_available:
+            return {
+                "Project": ["create_generic_asset"],
+                "Task": ["create_generic_asset"],
+            }
+
+        return {}
+
     ##############################################################################################################
     # helper methods which can be subclassed in custom hooks to fine tune the behaviour of things
 
--- a/hooks/tk-houdini_actions.py
+++ b/hooks/tk-houdini_actions.py
@@ -291,7 +291,7 @@ class HoudiniActions(HookBaseClass):
         if app.flowam_available:
             return {
                 "All": ["reference_copy_link", "download"],
-                "Houdini Workfile": ["open", "discard_draft"],
+                "Houdini Workfile": ["open", "reference_copy_link", "discard_draft"],
                 "Template": ["open", "download"],
             }
 
--- a/hooks/tk-houdini_actions.py
+++ b/hooks/tk-houdini_actions.py
@@ -292,6 +292,7 @@ class HoudiniActions(HookBaseClass):
             return {
                 "All": ["reference_copy_link", "download"],
                 "Houdini Workfile": ["open", "discard_draft"],
+                "Template": ["open", "download"],
             }
 
         return {}
--- a/hooks/tk-houdini_actions.py
+++ b/hooks/tk-houdini_actions.py
@@ -290,7 +290,7 @@ class HoudiniActions(HookBaseClass):
         app = self.parent
         if app.flowam_available:
             return {
-                "All": ["reference_copy_link"],
+                "All": ["reference_copy_link", "download"],
                 "Houdini Workfile": ["open", "discard_draft"],
             }
 
@@ -305,8 +305,8 @@ class HoudiniActions(HookBaseClass):
         app = self.parent
         if app.flowam_available:
             return {
-                "Project": ["create_generic_asset"],
-                "Task": ["create_generic_asset"],
+                "Project": ["build_new_template"],
+                "Task": ["build_new_scene"],
             }
 
         return {}
--- a/hooks/tk-maya_actions.py
+++ b/hooks/tk-maya_actions.py
@@ -337,6 +337,38 @@ class MayaActions(HookBaseClass):
         if name == "image_plane":
             self._create_image_plane(path, sg_publish_data)
 
+    def action_mappings(self) -> dict:
+        """
+        Returns the action mappings for the loader app.
+
+        :returns: Dictionary of action mappings.
+        """
+        app = self.parent
+        if app.flowam_available:
+            return {
+                "All": ["reference_copy_link", "download"],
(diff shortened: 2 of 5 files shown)


hooks/tk-houdini_actions.py
```diff
@@ -281,6 +281,36 @@ def execute_action(self, name, params, sg_publish_data):
...
+    def action_mappings(self) -> dict:
+        """
+        Returns the action mappings for the loader app.
+
+        :returns: Dictionary of action mappings.
+        """
+        app = self.parent
+        if app.flowam_available:
+            return {
+                "All": ["reference_copy_link"],
```
yungsiow: Missing "download" here I think?

hooks/tk-houdini_actions.py
```diff
@@ -281,6 +281,36 @@ def execute_action(self, name, params, sg_publish_data):
...
+        Returns the entity mappings for the loader app.
+
+        :returns: Dictionary of entity mappings.
+        """
+        app = self.parent
+        if app.flowam_available:
+            return {
+                "Project": ["create_generic_asset"],
+                "Task": ["create_generic_asset"],
+            }
```
yungsiow, line 311: Sorry bit confused about this one.  Shouldn't this only be available in the desktop app?
In Houdini, I assume we only want `build_new_scene `action on the Task.

hooks/tk-maya_actions.py
```diff
@@ -337,6 +337,38 @@ def execute_action(self, name, params, sg_publish_data):
...
+        Returns the action mappings for the loader app.
+
+        :returns: Dictionary of action mappings.
+        """
+        app = self.parent
+        if app.flowam_available:
+            return {
+                "All": ["reference_copy_link", "download"],
+                "Maya Workfile": ["reference_am", "open", "discard_draft"],
+                "Alembic": ["reference_am"],
```
yungsiow: Referencing isn't supported for Alembic as far as I know.  We can only offer "copy_reference_link" for this, so it can just fall into the All category.

hooks/tk-maya_actions.py
```diff
@@ -337,6 +337,38 @@ def execute_action(self, name, params, sg_publish_data):
...
+                "Template": ["open"],
```
yungsiow: Does this mean that Templates will include "reference_copy_link"?  Because I don't think we want it to.  That might give users the impression they're allowed to use templates as references and they aren't.

hooks/tk-maya_actions.py
```diff
@@ -337,6 +337,38 @@ def execute_action(self, name, params, sg_publish_data):
...
+        Returns the entity mappings for the loader app.
+
+        :returns: Dictionary of entity mappings.
+        """
+        app = self.parent
+        if app.flowam_available:
+            return {
+                "Project": ["create_generic_asset"],
+                "Task": ["create_generic_asset"],
+            }
```
yungsiow, line 373: Same comment as for tk-houdini.  This should be "build_new_scene" on the Task.  And on the Project, it should be "build_new_template" right?  Sorry I'm going by memory...

hooks/tk-nuke_actions.py
```diff
@@ -294,6 +294,48 @@ def execute_action(self, name, params, sg_publish_data):
...
+    def action_mappings(self) -> dict:
+        """
+        Returns the action mappings for the loader app.
+
+        :returns: Dictionary of action mappings.
+        """
+        app = self.parent
+        if app.flowam_available:
+            return {
+                "All": ["reference_copy_link"],
```
yungsiow: Missing "download" here I think.

hooks/tk-nuke_actions.py
```diff
@@ -294,6 +294,48 @@ def execute_action(self, name, params, sg_publish_data):
...
+        Returns the entity mappings for the loader app.
+
+        :returns: Dictionary of entity mappings.
+        """
+        app = self.parent
+        if app.flowam_available:
+            return {
+                "Project": ["create_generic_asset"],
+                "Task": ["create_generic_asset"],
+            }
```
yungsiow, line 340: Same here, needs "build_new_scene" at Task level, and "build_new_template" at Project level.

yungsiow (commented):
Hey Charlie, I tried to point out where changes need to be made, but not sure if I caught everything.  In general, I think all the settings need a sweep to ensure they follow these rules:

For DCC engines:
1. all file types allow "download" action
2. all non-template types allow "copy reference link"
3. all dcc types allow "open", "discard"
4. Maya types allow "reference" when in Maya
5. Generic types allow "create_read_node" when in Nuke
6. From a Task, user can "build new scene"
7. From a Project, user can "build new template"

For Desktop engine:
1. all file types allow "download action"
2. all non-template types allow "copy reference link"
3. all generic types allow "publish"
4. From a Task, user can "create new generic"
5. From a Project, user can "create new generic"

Hope this helps!

hooks/tk-maya_actions.py
carlos-villavicencio-adsk: Fortunately, an entry will entirely override the `All` entry. Not a merge of actions. To sum up, Templates won't include `reference_copy_link`.

chenm1adsk (approved):
LGTM

hooks/tk-maya_actions.py
yungsiow: In that case does that mean Maya Workfile and Alembic will not include "reference_copy_link" and "download", because they should :)

hooks/tk-houdini_actions.py
```diff
@@ -281,6 +281,37 @@ def execute_action(self, name, params, sg_publish_data):
...
+        """
+        Returns the action mappings for the loader app.
+
+        :returns: Dictionary of action mappings.
+        """
+        app = self.parent
+        if app.flowam_available:
+            return {
+                "All": ["reference_copy_link", "download"],
+                "Houdini Workfile": ["open", "discard_draft"],
```
yungsiow: Should also include reference_copy_link and download for Houdini workfile.

hooks/tk-maya_actions.py
```diff
@@ -337,6 +337,37 @@ def execute_action(self, name, params, sg_publish_data):
...
+        """
+        Returns the action mappings for the loader app.
+
+        :returns: Dictionary of action mappings.
+        """
+        app = self.parent
+        if app.flowam_available:
+            return {
+                "All": ["reference_copy_link", "download"],
+                "Maya Workfile": ["reference_am", "open", "discard_draft"],
```
yungsiow: Should also include "reference_copy_link" and "download" for Maya workfile.

hooks/tk-nuke_actions.py
```diff
@@ -294,6 +294,48 @@ def execute_action(self, name, params, sg_publish_data):
...
+        """
+        Returns the action mappings for the loader app.
+
+        :returns: Dictionary of action mappings.
+        """
+        app = self.parent
+        if app.flowam_available:
+            return {
+                "All": ["reference_copy_link", "download"],
+                "Nuke Workfile": ["open", "discard_draft"],
```
yungsiow: Should also include "reference_copy_link" and "download" for Nuke workfile.

yungsiow (approved):
Looks good, Charlie, thanks!

PR #154, opened 2026-07-21
SG-44439 Hide filter widgets on flowam mode as well

carlos-villavicencio-adsk:
Hide the following widget as it's not intended to be part of the loader in the first place.

image


chenm1adsk (approved):
LGTM

PR #156, opened 2026-07-21
SG-44442 Fix published files name search for flowam

carlos-villavicencio-adsk:
This pull request refines the filtering logic in `proxymodel_latestpublish.py` to ensure that search filtering is always applied before type filtering. This change improves the accuracy and predictability of the filter behavior, especially when both search and type filters are used.

Filtering logic improvements:

* The search filter is now applied before checking for type filters, ensuring that search results are always respected regardless of the type filter state.
* If the search filter is set but the item's searchable field is missing or empty, the item is now correctly excluded from the results.
* The code clarifies that when no type filter is active (`_valid_type_ids` is `None`), all items passing the search filter are accepted.


chenm1adsk (approved):
LGTM!

PR #157, opened 2026-07-22
SG-43959 UI/UX for VariantSet component selector

carlos-villavicencio-adsk:
## Description
This pull request adds comprehensive support for FlowAM variant container assets in the loader dialog, enabling users to select and interact with different asset variants directly from the UI. The main changes introduce a new variant selector widget, update the data models and dialog logic to recognize and handle variant containers, and ensure that the UI updates correctly when variant selections or async thumbnail loads occur.

**FlowAM Variant Container Support**

* Added a new `VariantSelectorWidget` (`flowam/variant_selector_widget.py`) that appears in the details panel when a FlowAM variant container asset is selected. This widget allows users to select among available variants, emitting signals to update the actions menu and history accordingly. [[1]](diffhunk://#diff-e237095509fe7126684765c4d51382734dd96c17bab2b66b751e06a09b65e7eeR1-R118) [[2]](diffhunk://#diff-6ce7f50bc61858ca8f8bfd954cfe0d1280521c84dbc2f170ff01da4399d3f031L1133-R1193)
* Updated the loader dialog (`dialog.py`) to detect when a selected asset is a variant container, display the variant selector, and route actions/history through the currently selected variant. The dialog also seeds the UI with the first variant by default. [[1]](diffhunk://#diff-6ce7f50bc61858ca8f8bfd954cfe0d1280521c84dbc2f170ff01da4399d3f031L1133-R1193) [[2]](diffhunk://#diff-6ce7f50bc61858ca8f8bfd954cfe0d1280521c84dbc2f170ff01da4399d3f031R1255-R1265)

**Model and Data Handling Enhancements**

* Modified `MedmLatestPublishModel` (`latestpublish_model.py`) to identify variant containers, store variant sets and prebuilt sg_data for each variant, and provide a helper to populate the model with a single card for the container asset. [[1]](diffhunk://#diff-1ed57a88e0d5321a3af194170169df8806da64d93205c065f8f50e4d9bc7d7b6R203-R229) [[2]](diffhunk://#diff-1ed57a88e0d5321a3af194170169df8806da64d93205c065f8f50e4d9bc7d7b6R330-R402)
* Added a new role (`VARIANT_SETS_ROLE`) to the model for storing variant set metadata.

**UI and Async Thumbnail Handling**

* Ensured that when a thumbnail for a history or publish item arrives asynchronously, the details panel updates the image if that item is currently selected. This includes new signal connections and handlers for both history and publish models. [[1]](diffhunk://#diff-6ce7f50bc61858ca8f8bfd954cfe0d1280521c84dbc2f170ff01da4399d3f031R186-R191) [[2]](diffhunk://#diff-6ce7f50bc61858ca8f8bfd954cfe0d1280521c84dbc2f170ff01da4399d3f031R2702-R2704) [[3]](diffhunk://#diff-6ce7f50bc61858ca8f8bfd954cfe0d1280521c84dbc2f170ff01da4399d3f031R1276-R1407)

**Imports and Initialization**

* Updated imports to include the new `VariantSelectorWidget` in the FlowAM module initialization.
* Added initialization and cleanup logic for the variant selector widget in the dialog.

## Screenshots

image

image

### DIFF
--- a/python/tk_multi_loader/dialog.py
+++ b/python/tk_multi_loader/dialog.py
@@ -135,6 +135,11 @@ class AppDialog(QtGui.QWidget):
         # the GC happy.
         self._dynamic_widgets = []
 
+        # Flag used to prevent re-entrant calls to _popup_menu.
+        # A spurious context menu event can be generated when a modal
+        # dialog opened from a menu action closes (e.g. Create New template).
+        self._popup_menu_active = False
+
         # maintain a special flag so that we can switch profile
         # tabs without triggering events
         self._disable_tab_event_handler = False
@@ -1914,13 +1919,38 @@ class AppDialog(QtGui.QWidget):
 
         :param position: The position where the menu should be displayed.
         """
+        # Guard against re-entrant calls. When an action opens a modal dialog,
+        # closing that dialog can cause Qt to deliver a spurious context menu
+        # event (e.g. a replayed WM_CONTEXTMENU on Windows). Without this guard
+        # the menu would re-appear as soon as the dialog is dismissed.
+        if self._popup_menu_active:
+            return
+
         view = self.sender()
-        menu = QtGui.QMenu(view)
 
-        for action in view.actions():
-            menu.addAction(action)
+        # Do not show the context menu when right-clicking on an empty area.
+        if not view.indexAt(position).isValid():
+            return
 
-        menu.exec_(view.viewport().mapToGlobal(position))
+        self._popup_menu_active = True
+        try:
+            menu = QtGui.QMenu(view)
+
+            for action in view.actions():
+                menu.addAction(action)
+
+            menu.exec_(view.viewport().mapToGlobal(position))
+        finally:
+            # Defer the flag reset so the guard stays active for one additional
+            # event-loop iteration. This discards any spurious ContextMenu event
+            # that was queued during the menu or dialog execution (e.g. an async
+            # file-open triggered by an action that eventually generates a
+            # WM_CONTEXTMENU message for the parent window).
+            QtCore.QTimer.singleShot(0, self._clear_popup_menu_flag)
+
+    def _clear_popup_menu_flag(self):
+        """Reset the popup menu active flag after pending events are processed."""
+        self._popup_menu_active = False
 
     def _set_contextual_menu(
         self, sg_data: dict, field_value: Any, view: Any, model: Any
@@ -2020,6 +2050,10 @@ class AppDialog(QtGui.QWidget):
         # ---------------------------------------------------------------
 
         view.setContextMenuPolicy(QtCore.Qt.CustomContextMenu)
+        try:
+            view.customContextMenuRequested.disconnect(self._popup_menu)
+        except RuntimeError:
+            pass
         view.customContextMenuRequested.connect(self._popup_menu)
 
     def _get_entity_root(self, root):
--- a/python/tk_multi_loader/dialog.py
+++ b/python/tk_multi_loader/dialog.py
@@ -383,6 +383,11 @@ class AppDialog(QtGui.QWidget):
             # Disable filter menu for Flow Asset Management mode - it expects ShotgunModel data
             self._filter_menu = None
             self.ui.filter_menu_btn.hide()
+            # Also hide the legacy filter widgets
+            self.ui.publish_type_list.hide()
+            self.ui.publish_type_filter_title.hide()
+            self.ui.check_all.hide()
+            self.ui.check_none.hide()
         else:
             # Hide the legacy filter widgets
             self.ui.publish_type_list.hide()
--- a/python/tk_multi_loader/dialog.py
+++ b/python/tk_multi_loader/dialog.py
@@ -1168,8 +1168,10 @@ class AppDialog(QtGui.QWidget):
                     # When the user picks a different variant, refresh the
                     # actions menu for that variant.
                     selector.selection_changed.connect(
-                        lambda sn, vn, aid, vsd=variant_sg_dicts: (
-                            self._update_actions_for_variant(aid, vsd)
+                        lambda _set_name, _variant_name, asset_id, variant_set_dicts=variant_sg_dicts: (
+                            self._update_actions_for_variant(
+                                asset_id, variant_set_dicts
+                            )
                         )
                     )
 
--- a/python/tk_multi_loader/dialog.py
+++ b/python/tk_multi_loader/dialog.py
@@ -93,6 +93,11 @@ class AppDialog(QtGui.QWidget):
         # FlowAM tree view - only created when FlowAM is enabled
         self._medm_tree_view = None
 
+        # Variant selector widget shown in the details panel when a FlowAM
+        # variant container asset is selected.  Created on first use and
+        # re-populated on every selection change.
+        self._medm_variant_selector = None
+
         # The loader app can be invoked from other applications with a custom
         # action manager as a File Open-like dialog. For these managers, we won't
         # be using the banner system.
@@ -176,6 +181,12 @@ class AppDialog(QtGui.QWidget):
             # FlowAM history model for FlowAM publish items
             self._medm_history_model = MedmPublishHistoryModel(
                 self, self._task_manager, self._medm_cache, self._medm_thumbnail_service
+            )
+
+            # When a history item's thumbnail arrives asynchronously, re-apply
+            # it to details_image if that item is currently selected.
+            self._medm_history_model.dataChanged.connect(
+                self._on_history_item_data_changed
             )
 
         self._publish_history_model_overlay = ShotgunModelOverlayWidget(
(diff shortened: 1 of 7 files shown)


python/tk_multi_loader/flowam/variant_selector_widget.py
```diff
@@ -0,0 +1,118 @@
...
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Software Inc. License Agreement included in this
+# distribution package. See LICENSE.
+
+"""Variant selector widget for FlowAM variant container assets.
```
chenm1adsk: Do we need this file docstring since there is only one class in this file?

python/tk_multi_loader/flowam/variant_selector_widget.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class VariantSelectorWidget(QtGui.QWidget):
+    """Widget that renders one labelled row per variant set axis.
+
+    Each row shows the set name as a label and a :class:`QComboBox` listing
+    the available variant options.  Selecting a different option emits
+    :attr:`selection_changed`.
+
+    :param variant_sets: Mapping of ``set_name -> [(variant_name, asset_id)]``.
```
chenm1adsk: I am wondering if we are going to switch docstring style from Sphinx to google style?
Also can we move these two input params description to the __init__() function instead?

python/tk_multi_loader/flowam/variant_selector_widget.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def get_selected_asset_id(self, set_name: str) -> Optional[str]:
+        """Return the currently selected asset_id for *set_name*, or ``None``."""
+        combo = self._combos.get(set_name)
+        return combo.currentData() if combo else None
+
+    def get_all_selections(self) -> Dict[str, Tuple[str, str]]:
+        """Return ``{set_name: (variant_name, asset_id)}`` for every set."""
+        return {
+            sn: (cb.currentText(), cb.currentData()) for sn, cb in self._combos.items()
```
chenm1adsk: can we use set_name instead of sn? (Sorry I am a hater of abbreviation 😢 )

python/tk_multi_loader/flowam/variant_selector_widget.py
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+    # ------------------------------------------------------------------
+    # Private – UI construction
+    # ------------------------------------------------------------------
+
+    def _build_ui(self) -> None:
```
chenm1adsk, line 81: Should we consider separate model/view code? Would it be overkill?

python/tk_multi_loader/flowam/latestpublish_model.py
```diff
@@ -297,6 +327,79 @@ def _populate_model_from_selected_item(
...
+        asset: objects.FlowAsset,
+        variant_sets: dict[str, list[tuple[str, str]]],
+    ) -> None:
+        """Show *asset* as a single center-panel card for a variant container.
+
+        The card carries two extra keys in its sg_data:
+
+        * ``_variant_sets``: the raw variant-sets dict
+          ``{set_name: [(variant_name, asset_id)]}``
+        * ``_variant_sg_dicts``: pre-built sg_data dicts keyed by *asset_id*,
```
chenm1adsk: This part of the docstring is confusing. The _variant_sg_dicts is built inside this function right? Can you be more clear?

python/tk_multi_loader/flowam/latestpublish_model.py
```diff
@@ -297,6 +327,79 @@ def _populate_model_from_selected_item(
...
+          one per variant cell.  These are used by the details panel to
+          populate the actions menu when the user picks a variant.
+
+        :param asset: The container :class:`FlowAsset`.
+        :param variant_sets: Mapping returned by :meth:`ComponentMixin.get_variant_sets`.
+        """
+        # Collect every unique variant-cell asset_id across all sets.
+        all_variant_asset_ids: set[str] = set()
+        for variants in variant_sets.values():
+            for _vname, asset_id in variants:
```
chenm1adsk: asset_id here refers to the asset_id that the variant set component points to, which is easy to confuse with the asset parameter above. Should we consider use something like variant_set_asset_id for clarity? (I know it is a bad name too)

python/tk_multi_loader/flowam/latestpublish_model.py
```diff
@@ -297,6 +327,79 @@ def _populate_model_from_selected_item(
...
+                if asset_id:
+                    all_variant_asset_ids.add(asset_id)
+
+        # Fetch the variant cell FlowAsset objects (best-effort; use cache).
+        variant_sg_dicts: dict[str, dict[str, Any]] = {}
+        try:
+            if asset.id in self._cache.children:
```
chenm1adsk: can we make this block a separate helper function? I remember similar logic also applied elsewhere

python/tk_multi_loader/flowam/latestpublish_model.py
```diff
@@ -297,6 +327,79 @@ def _populate_model_from_selected_item(
         sg_publish_type_counts = self._calculate_sg_publish_type_counts()
         self._publish_type_model.set_active_types(sg_publish_type_counts)
 
+    def _populate_variant_container(
```
chenm1adsk, line 330: Not saying it has to be done in the scope of this PR but this function is growing. We should consider refactor it at some point to something like:


_populate_model_from_selected_item(selected_item)
│
├── _resolve_asset(selected_item) -> Optional[FlowAsset]
│     └── extract + validate, return None if invalid
│
├── [early return] if variant container:
│     └── _populate_variant_container(asset, variant_sets)  # already extracted
│
├── _resolve_leaf_assets(asset) -> list[sg_dict]
│     ├── _fetch_asset_children(asset)
│     └── fallback: [_asset_to_sg_dict(asset)] if leaf, else []
│
├── _fetch_drafts_by_asset_id(assets) -> dict[asset_id, list[sg_dict]]
│     └── per-asset draft fetch + cache + convert
│
└── _emit_center_panel_cards(children_sg_dicts, drafts_by_asset_id, leaf_fallback, parent_asset_id)
      ├── published vs draft priority logic (Pass 2)
      ├── leaf_fallback draft surface (edge case)
      └── new draft items (Pass 3)

python/tk_multi_loader/dialog.py
```diff
@@ -1130,26 +1141,67 @@ def __clear_publish_history(pixmap):
...
-                        self._details_action_menu.addAction(a)
+                    # Normal (non-variant) publish: restore history view and
+                    # build the actions menu from the item's own sg_data.
+                    self.ui.history_view.setEnabled(True)
+
+                    actions = self._action_manager.get_actions_for_publish(
+                        sg_item, self._action_manager.UI_AREA_DETAILS
+                    )
+                    if len(actions) == 0:
+                        self.ui.detail_actions_btn.setVisible(False)
```
chenm1adsk, line 1211: Do we need to set self.ui.detail_actions_btn.setVisible(False) back to True in else branch?

python/tk_multi_loader/dialog.py
```diff
@@ -1130,26 +1141,67 @@ def __clear_publish_history(pixmap):
...
+                    selector = VariantSelectorWidget(
+                        variant_sets, self.ui.details_widget
+                    )
+                    self._medm_variant_selector = selector
+                    self.ui.details_widget.layout().addWidget(selector)
+
+                    # When the user picks a different variant, refresh the
+                    # actions menu for that variant.
+                    selector.selection_changed.connect(
+                        lambda sn, vn, aid, vsd=variant_sg_dicts: (
```
chenm1adsk: this lambda function a bit complicated to read. Can we make it easier?

python/tk_multi_loader/dialog.py
```diff
@@ -1130,26 +1141,67 @@ def __clear_publish_history(pixmap):
...
+
+                    # Clear stale history before seeding with the first variant.
+                    self._publish_history_model.clear()
+                    if self._medm_history_model is not None:
+                        self._medm_history_model.clear()
+                    self._update_history_view_height()
+
+                    # Seed the actions menu and history with the first option of
+                    # the first set so the panel is immediately usable.
+                    first_set_variants = next(iter(variant_sets.values()), [])
```
chenm1adsk: can we move this first variant seed logic inside VariantSelectorWidget.__init__?

chenm1adsk (commented):
Left some comments and questions!

python/tk_multi_loader/flowam/variant_selector_widget.py
carlos-villavicencio-adsk: We'll definitely stick to Sphinx, which we use to generate the current documentation. Good catch.

python/tk_multi_loader/flowam/latestpublish_model.py
carlos-villavicencio-adsk: I can't think of a better name suggestion. Thanks!

python/tk_multi_loader/flowam/variant_selector_widget.py
carlos-villavicencio-adsk: All I can recognize is UI view code. Can you point me to

python/tk_multi_loader/dialog.py
carlos-villavicencio-adsk: No need. It's being called beforehand. Placing it there will be redundant.

python/tk_multi_loader/dialog.py
carlos-villavicencio-adsk: I did some refactors here so the code is readable and now it depends on `VariantSelectorWidget.get_selected_asset_id` method instead.

python/tk_multi_loader/flowam/variant_selector_widget.py
chenm1adsk: I actually don't have enough knowledge about QT at all. 😢

chenm1adsk (approved):
Looks good to me!

PR #149, opened 2026-07-14
SG-43935 Update creation of dcc workfile (data model base)

yungsiow:
NOTE: this branch will be treated as the base branch for publisher changes involving the Data Model project

- add .flake8 copied from tk-core
- add vim files to .gitignore
- remove restrictions of single dcc workfile of same type per parent
- add workfile type designation to workfile asset name
- remove sg_task_name parameter
- fix flake complaints

PR #160, opened 2026-08-05
SG-44680 Prevent downloading and copy ref link for container assets

carlos-villavicencio-adsk:
Base branch: #149 

Adds validation to prevent displaying "download" and "copy reference link" for asset containers (root assets) Introduced on the component.VariantSet data model.

### DIFF
--- a/hooks/tk-desktop_actions.py
+++ b/hooks/tk-desktop_actions.py
@@ -82,9 +82,7 @@ class DesktopActions(HookBaseClass):
             if (
                 "download" in actions
                 and sg_publish_data.get("type") == "PublishedFile"
-                and not flowam_actions.is_root_asset(
-                    sg_publish_data.get("_medm_asset")
-                )
+                and not flowam_actions.is_root_asset(sg_publish_data.get("_medm_asset"))
             ):
                 version_number = sg_publish_data.get("version_number")
 
@@ -132,9 +130,7 @@ class DesktopActions(HookBaseClass):
                     "version_number", flowam_actions.DRAFT_VERSION_IDENTIFIER
                 )
                 > flowam_actions.DRAFT_VERSION_IDENTIFIER
-                and not flowam_actions.is_root_asset(
-                    sg_publish_data.get("_medm_asset")
-                )
+                and not flowam_actions.is_root_asset(sg_publish_data.get("_medm_asset"))
             ):
                 action_instances.append(
                     {
--- a/hooks/tk-desktop_actions.py
+++ b/hooks/tk-desktop_actions.py
@@ -79,7 +79,13 @@ class DesktopActions(HookBaseClass):
         if app.flowam_available:
             flowam_actions = app.flowam.FlowAMActions()
 
-            if "download" in actions and sg_publish_data.get("type") == "PublishedFile":
+            if (
+                "download" in actions
+                and sg_publish_data.get("type") == "PublishedFile"
+                and not flowam_actions.is_container_asset(
+                    sg_publish_data.get("_medm_asset")
+                )
+            ):
                 version_number = sg_publish_data.get("version_number")
 
                 if (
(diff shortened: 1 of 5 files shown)


hooks/tk-desktop_actions.py
```diff
@@ -79,7 +79,13 @@ def generate_actions(
         if app.flowam_available:
             flowam_actions = app.flowam.FlowAMActions()
 
-            if "download" in actions and sg_publish_data.get("type") == "PublishedFile":
+            if (
```
chenm1adsk, line 82: Are we support any action from middle panel asset cards? If no should we just remove the action item from there?

carlos-villavicencio-adsk: Sounds like a good idea. However, this logic is evaluated on any item on all panels. So the only way to differenciate them if to add this kind of condition.

hooks/tk-maya_actions.py
```diff
@@ -316,6 +322,7 @@ def execute_action(self, name, params, sg_publish_data):
                 flowam_actions._build_new_template(sg_publish_data)
 
             if name == "download":
+                print("@@@@@", sg_publish_data.get("_medm_asset"))
```
yungsiow: leftover print statement.

python/tk_multi_loader/flowam/flowam_actions.py
```diff
@@ -402,6 +407,21 @@ def _get_dialog_parent(self) -> QtGui.QWidget | None:
...
+        """
+        Determine if the given FlowAsset is a container asset.
+
+        :param asset: The FlowAsset to check.
+        :returns: True if the asset is a container, False otherwise.
+        """
+        if (
+            asset.find_component(type_id=schema.get_schema_id(CONTAINER_TYPE))
+            is not None
+        ):
```
yungsiow: I think we could just remove the if statement here and just return the value of find_component() is not None right?

python/tk_multi_loader/flowam/flowam_actions.py
```diff
@@ -416,7 +436,17 @@ def _download_asset_revision(self, sg_publish_data: dict) -> None:
...
+            result = download_revision(flow_revision_id)
+        except DownloadRevisionError as exc:
+            if 'Component of purpose "source" does not exist on revision.' in str(exc):
+                QtGui.QMessageBox.warning(
+                    None,
+                    "Warning",
+                    "This asset has no binaries to download.",
+                )
+                return
+            raise
```
yungsiow, line 443: I'm assuming we can't actually get here now that we've added the download filter on the actions.  This is just for extra safety?

python/tk_multi_loader/flowam/flowam_actions.py
yungsiow: Also, I wonder if we should actually call the function "is_root_asset" since that's actually more accurately what we're looking for, and I think in the future, we will type the root asset explicitly as that type. (though I'm not sure about that).

python/tk_multi_loader/flowam/flowam_actions.py
carlos-villavicencio-adsk: Yes, extra safety. This was the first iteration of the fix. But I don't think it will hurt. Do you think it's better to remove it?

python/tk_multi_loader/flowam/flowam_actions.py
```diff
@@ -26,7 +26,12 @@
     create_dcc_workfile,
     create_template_workfile,
 )
-from .file import open_draft, download_revision, checkout_revision
+from .file import (
+    DownloadRevisionError,
+    open_draft,
+    download_revision,
+    checkout_revision,
```
yungsiow, line 31: let's maintain alphabetical order here.

python/tk_multi_loader/flowam/flowam_actions.py
yungsiow: No I like it - just in case!

yungsiow (approved):
Personally I would ignore case when alphabetizing, but not a huge deal.  Looks good, thanks Charlie!

